PR #5, opened 2019-10-08
rotate expired wrangler cert

galal-hussein:
https://github.com/rancher/k3s/issues/875

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -36,6 +36,8 @@ import (
 	"path"
 	"strings"
 	"time"
+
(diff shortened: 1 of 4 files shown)


erikwilson:
Thanks for the quick PR! 

It would be nice if we could avoid duplicating the cert expiration logic as it already exists in `server.go`. The check happening here could be moved to `storage.go` just before `server.Update`, or dynamiclistener configuration could take a generic function for rejecting/accepting a cert, or dynamiclistener could be the authority for the expiration check and `server.go` updated to call a function from this package.

PR #28, opened 2020-08-05
Allow forcing cert reissuance

brandond:
Refreshing the cert should force reissuance as opposed to returning
early if the SANs aren't changing. This is currently breaking refresh
of expired certs as per:
https://github.com/rancher/k3s/issues/1621#issuecomment-669464318

Signed-off-by: Brad Davidson <brad.davidson@rancher.com>

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -126,7 +126,7 @@ func NewSignedCert(cfg Config, key crypto.Signer, caCert *x509.Certificate, caKe
 
 	parsedCert, err := x509.ParseCertificate(certDERBytes)
 	if err == nil {
-		logrus.Infof("certificate %v signed by %v: notBefore=%v notAfter=%v",
+		logrus.Infof("certificate %s signed by %s: notBefore=%s notAfter=%s",
 			parsedCert.Subject, caCert.Subject, parsedCert.NotBefore, parsedCert.NotAfter)
 	}
 	return parsedCert, err
@@ -285,7 +285,7 @@ func IsCertExpired(cert *x509.Certificate, days int) bool {
 	expirationDate := cert.NotAfter
 	diffDays := time.Until(expirationDate).Hours() / 24.0
 	if diffDays <= float64(days) {
-		logrus.Infof("certificate %v will expire in %f days at %v", cert.Subject, diffDays, cert.NotAfter)
+		logrus.Infof("certificate %s will expire in %f days at %s", cert.Subject, diffDays, cert.NotAfter)
 		return true
 	}
 	return false
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -45,6 +45,10 @@ const (
 	duration365d = time.Hour * 24 * 365
 )
 
+var (
+	ErrStaticCert = errors.New("cannot renew static certificate")
+)
+
 // Config contains the basic fields required for creating a certificate
 type Config struct {
 	CommonName   string
@@ -119,7 +123,13 @@ func NewSignedCert(cfg Config, key crypto.Signer, caCert *x509.Certificate, caKe
 	if err != nil {
 		return nil, err
 	}
-	return x509.ParseCertificate(certDERBytes)
+
+	parsedCert, err := x509.ParseCertificate(certDERBytes)
+	if err == nil {
+		logrus.Infof("certificate %v signed by %v: notBefore=%v notAfter=%v",
+			parsedCert.Subject, caCert.Subject, parsedCert.NotBefore, parsedCert.NotAfter)
+	}
+	return parsedCert, err
 }
 
 // MakeEllipticPrivateKeyPEM creates an ECDSA private key
@@ -271,11 +281,11 @@ func ipsToStrings(ips []net.IP) []string {
 }
 
 // IsCertExpired checks if the certificate about to expire
-func IsCertExpired(cert *x509.Certificate) bool {
+func IsCertExpired(cert *x509.Certificate, days int) bool {
 	expirationDate := cert.NotAfter
-	diffDays := expirationDate.Sub(time.Now()).Hours() / 24.0
-	if diffDays <= 90 {
-		logrus.Infof("certificate will expire in %f days", diffDays)
+	diffDays := time.Until(expirationDate).Hours() / 24.0
+	if diffDays <= float64(days) {
+		logrus.Infof("certificate %v will expire in %f days at %v", cert.Subject, diffDays, cert.NotAfter)
 		return true
 	}
 	return false
--- a/factory/cert_utils.go
+++ b/factory/cert_utils.go
@@ -96,7 +96,7 @@ func NewSignedCert(signer crypto.Signer, caCert *x509.Certificate, caKey crypto.
 
 	parsedCert, err := x509.ParseCertificate(cert)
 	if err == nil {
-		logrus.Infof("certificate %v signed by %v: notBefore=%v notAfter=%v",
+		logrus.Infof("certificate %s signed by %s: notBefore=%s notAfter=%s",
 			parsedCert.Subject, caCert.Subject, parsedCert.NotBefore, parsedCert.NotAfter)
 	}
 	return parsedCert, err
--- a/factory/cert_utils.go
+++ b/factory/cert_utils.go
@@ -11,6 +11,8 @@ import (
 	"math/big"
 	"net"
 	"time"
+
+	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -92,7 +94,12 @@ func NewSignedCert(signer crypto.Signer, caCert *x509.Certificate, caKey crypto.
 		return nil, err
 	}
 
-	return x509.ParseCertificate(cert)
+	parsedCert, err := x509.ParseCertificate(cert)
+	if err == nil {
+		logrus.Infof("certificate %v signed by %v: notBefore=%v notAfter=%v",
+			parsedCert.Subject, caCert.Subject, parsedCert.NotBefore, parsedCert.NotAfter)
+	}
+	return parsedCert, err
 }
 
 func ParseCertPEM(pemCerts []byte) (*x509.Certificate, error) {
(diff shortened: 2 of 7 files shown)


factory/gen.go
```diff
@@ -91,16 +91,20 @@ func (t *TLS) Filter(cn ...string) []string {
...
 	if !NeedsUpdate(0, secret, cn...) {
 		return secret, false, nil
 	}
 
+	return t.generateCert(secret, cn...)
+}
+
+func (t *TLS) generateCert(secret *v1.Secret, cn ...string) (*v1.Secret, bool, error) {
+	var (
+		err error
```
briandowns, line 105: Is this necessary?

cert/cert.go
```diff
@@ -45,6 +45,10 @@ const (
 	duration365d = time.Hour * 24 * 365
 )
 
+var (
+	ErrStaticCert = errors.New("cannot renew static certificate")
```
briandowns, line 49: Is this value used outside of this package?

factory/cert_utils.go
```diff
@@ -92,7 +94,12 @@ func NewSignedCert(signer crypto.Signer, caCert *x509.Certificate, caKey crypto.
 		return nil, err
 	}
 
-	return x509.ParseCertificate(cert)
+	parsedCert, err := x509.ParseCertificate(cert)
+	if err == nil {
+		logrus.Infof("certificate %v signed by %v: notBefore=%v notAfter=%v",
```
briandowns, line 99: If any of the values being printed are supported, ie %s, %d, etc, the format should be updated accordingly.

cert/cert.go
```diff
@@ -119,7 +123,13 @@ func NewSignedCert(cfg Config, key crypto.Signer, caCert *x509.Certificate, caKe
 	if err != nil {
 		return nil, err
 	}
-	return x509.ParseCertificate(certDERBytes)
+
+	parsedCert, err := x509.ParseCertificate(certDERBytes)
+	if err == nil {
+		logrus.Infof("certificate %v signed by %v: notBefore=%v notAfter=%v",
```
dweomer, line 129: Is this for debugging primarily or is it something that we would expect user/operators to dig into their logs looking for?

factory/gen.go
```diff
@@ -67,40 +65,51 @@ func collectCNs(secret *v1.Secret) (domains []string, ips []net.IP, hash string,
 		}
 	}
 
-	hash = hex.EncodeToString(digest.Sum(nil))
 	return
 }
 
 func (t *TLS) Merge(target, additional *v1.Secret) (*v1.Secret, bool, error) {
-	return t.AddCN(target, cns(additional)...)
+	secret, updated, err := t.AddCN(target, cns(additional)...)
```
briandowns, line 72: This would make a great doc comment for the function itself and update this with this code intent.

listener.go
```diff
@@ -22,7 +23,7 @@ type TLSStorage interface {
 }
 
 type TLSFactory interface {
-	Refresh(secret *v1.Secret) (*v1.Secret, error)
+	Renew(secret *v1.Secret) (*v1.Secret, error)
```
dweomer, line 26: Would like to understand the reasoning behind changing this exported interface method name. This will have the effect of breaking existing clients when they attempt to upgrade (which may be on purpose, I don't know).

listener.go
```diff
@@ -152,13 +153,13 @@ type listener struct {
 func (l *listener) WrapExpiration(days int) net.Listener {
 	ctx, cancel := context.WithCancel(context.Background())
 	go func() {
-		time.Sleep(5 * time.Minute)
+		time.Sleep(30 * time.Second)
```
briandowns, line 156: Can you include a comment as to why we're sleeping here and the significance of 30 seconds?

factory/gen.go
```diff
@@ -133,7 +142,7 @@ func (t *TLS) AddCN(secret *v1.Secret, cn ...string) (*v1.Secret, bool, error) {
 	}
 	secret.Data[v1.TLSCertKey] = certBytes
 	secret.Data[v1.TLSPrivateKeyKey] = keyBytes
-	secret.Annotations[hashKey] = hash
+	secret.Annotations[hashKey] = fmt.Sprintf("SHA1=%X", sha1.Sum(newCert.Raw))
```
dweomer, line 145: Why have we down-graded the hash algo?

listener.go
```diff
@@ -366,6 +364,15 @@ func (l *listener) loadCert() (*tls.Certificate, error) {
...
 
+	// cert has changed, close closeWrapper wrapped connections
+	if l.conns != nil {
+		l.connLock.Lock()
+		for _, conn := range l.conns {
+			_ = conn.close()
+		}
+		l.connLock.Unlock()
+	}
+
```
dweomer, line 375: So, prior to this change, connections were reset when calling `updateCert()` but with this change it is on `loadCert()` which leads to different behavior when calling `getCertificate()`. Is this intentional?

listener.go
brandond: I am not sure what the original intent was in sleeping for 5 minutes at startup before checking certificate renewal. I have not seen any particular reason for it based on how dynamiclistener is used in k3s and rancher. @ibuildthecloud or @galal-hussein may have some context?

In practice, it makes it harder to test the fix.

factory/gen.go
brandond: Previously this was a SHA256 hash of not the certificate itself, but the SubjectAlternativeNames attribute text. Since we need to detect and handle changes to the cert as a whole in order to invalidate the caches, we are now hashing all of it which makes this a fingerprint. SHA1 is the standard fingerprint alg for other tools such as `openssl x509`, the Windows certificate management UI, and OS X's Keychain, so using the same alg here makes it easy to compare the certificate information across tools.

listener.go
brandond: There was a defect with the connection reset code, where only the node updating the certificate would reset its connections. The other nodes would load the new cert from the k8s store, but not close any connections, leaving them on the old cert. This resolves that.

listener.go
brandond: The industry standard term for what we are doing here is renewal, not refreshing. See:
https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.0-1.pdf section 4.6. Using standard terminology makes it clear what this method is responsible for doing.

cert/cert.go
brandond: We already log the expiration at info level, so I thought it would be nice to also log the issuance so that anyone looking at the logs knows that the expiration has been extended. Do you think this (and the expiration) should be debug?

factory/cert_utils.go
brandond: these are pkix.Name and time.Time. Is it preferred to use %s for non-string types or let %v handle the formatting?

cert/cert.go
brandond: it is used outside dynamiclistener/cert, in dynamiclistener/factory and dynamiclistener, but not outside this module.

factory/cert_utils.go
briandowns: It's better to be explicit whenever possible.

cert/cert.go
dweomer: It's hard to say. With logs I try to err on the side of minimal logs in "prod" aka the default which tends to route stuff to debug. I think with this change you have a number of infos that should probably be debugs however it would be valuable and informative to print this to info once per event.

factory/gen.go
dweomer: Would it make sense to have a `const fingerprint = "listener.cattle.io/hash"` and use this as the annotation key instead of re-using the old key which had different semantics?

listener.go
dweomer: > The industry standard term for what we are doing here is renewal, not refreshing. See:
> https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.0-1.pdf section 4.6. Using standard terminology makes it clear what this method is responsible for doing.

Sorry, I assumed this. I agree it is a better name but are there consequences that we might wish to avoid with such a breaking change in the public API. Yes we own it. I guess what I am asking is, is it only ever "internally" consumed? If so, fine. If not, are we comfortable breaking downstream?

listener.go
dweomer: Seems like a decent balance.

factory/gen.go
brandond: I am 100% down with changing the annotation key. I checked k3s and rancher and it didn't appear that anything else was relying on the annotation, so changing it here should be fine.

listener.go
brandond: I don't see any use of the Renew method in Rancher or K3s; it appears to only be called internal to this module. Given that we don't even have a README in this repo, I somewhat doubt anyone else is using this module at the moment, but I very well could be wrong.

factory/gen.go
brandond: I added docs to the other exported functions in this method as well, hopefully that explains the flow a little better?

cert/cert.go
brandond: Our embedded certificate stuff feels like a bit of a black box at the moment. These logs should only trigger when the cluster is first created, and then again 9 months later when the certs are a about to expire, so these should not be very chatty despite being at info level.

factory/gen.go
```diff
@@ -67,40 +65,59 @@ func collectCNs(secret *v1.Secret) (domains []string, ips []net.IP, hash string,
 		}
 	}
 
-	hash = hex.EncodeToString(digest.Sum(nil))
 	return
 }
 
+// Merge combines the SAN lists from the target and additional Secrets, and returns a potentially modified Secret,
```
briandowns, line 71: I love all of the doc comments you added. Thank you very much.

listener.go
```diff
@@ -152,13 +153,13 @@ type listener struct {
...
 	go func() {
-		time.Sleep(5 * time.Minute)
+		time.Sleep(30 * time.Second)
 
 		for {
 			wait := 6 * time.Hour
-			if err := l.checkExpiration(days); err != nil {
-				logrus.Errorf("failed to check and refresh dynamic cert: %v", err)
-				wait = 5 + time.Minute
+			if err := l.checkExpiration(days); err != nil && err != cert.ErrStaticCert {
```
briandowns, line 160: Can you include a comment describing the state of when `l.checkExpiration` returns an error that IS `cert.ErrStaticCert`?

brandond: I added a comment. Given that there's user action necessary to resolve this state, I think continuing to print errors to the log. We could also consider attaching an Event to the Kubernetes secret that the cert is stored in, but that's probably out of scope for resolving this issue.

PR #29, opened 2020-08-05
Allow forcing cert reissuance (v0.2 backport)

brandond:
Backport of #28 for the release-0.2 branch. Also pulls in a few recent fixes from master.

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -45,6 +45,10 @@ const (
 	duration365d = time.Hour * 24 * 365
 )
 
+var (
+	ErrStaticCert = errors.New("cannot renew static certificate")
+)
+
 // Config contains the basic fields required for creating a certificate
 type Config struct {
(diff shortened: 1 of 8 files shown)


cert/cert.go
```diff
@@ -271,11 +281,11 @@ func ipsToStrings(ips []net.IP) []string {
 }
 
 // IsCertExpired checks if the certificate about to expire
-func IsCertExpired(cert *x509.Certificate) bool {
+func IsCertExpired(cert *x509.Certificate, days int) bool {
```
dweomer, line 284: IMO, this should be a new function called `IsCertExpiring(cert,days)`

brandond: I agree, it probably should have been.

brandond:
I just rebased this with the final commit that was merged to master.

PR #35, opened 2021-04-23
Update IsStatic to check for nil annotations

dramich:


brandond (approved):
yay nil maps

factory/gen.go
```diff
@@ -179,7 +179,10 @@ func populateCN(secret *v1.Secret, cn ...string) *v1.Secret {
 // IsStatic returns true if the Secret has an attribute indicating that it contains
 // a static (aka user-provided) certificate, which should not be modified.
 func IsStatic(secret *v1.Secret) bool {
-	return secret.Annotations[Static] == "true"
+	if secret.Annotations != nil {
+		return secret.Annotations[Static] == "true"
+	}
+	return false
```
brandond, line 185:
```suggestion
	return secret.Annotations != nil && secret.Annotations[Static] == "true"
```

PR #36, opened 2021-05-03
prepare expiry support for utils

mschneider82:
https://github.com/k3s-io/k3s/issues/3253

currently the expiry date values are hard-coded and cannot be configured from external. I have added additional functions to support custom valid until values for cert_utils used by k3s.


xiaods:
current codebase already fix it. you can use upstream code to resove your issue.

PR #41, opened 2021-07-12
Merge TLS only if TLS factory is set

weihanglo:
Since `storage.tls` is optional, we should check it existence before calling its methods.


guangbochen (approved):
LGTM

PR #43, opened 2021-10-26
Add ability to force cert regeneration

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>

### DIFF
--- a/factory/gen.go
+++ b/factory/gen.go
@@ -119,6 +119,11 @@ func (t *TLS) AddCN(secret *v1.Secret, cn ...string) (*v1.Secret, bool, error) {
 	return t.generateCert(secret, cn...)
 }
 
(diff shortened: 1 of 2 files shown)


brandond:
Just want to be clear in the terminology used here about regenerating new certs vs renewing them. It is very important to Rancher that the cert is just **renewed** and not actually regenerated, as generating a new cert makes the browser suspicious and will break websockets.

briandowns:
This operation would generate a new certificate needed for the certificate rotation work.

galal-hussein (approved):

brandond (approved):
OK, my apologies. This is not intended to be used to force renewal; the expected behavior is to repopulate the secret with a completely new key and certificate, retaining only the CNs from the current certificate.

PR #44, opened 2021-10-29
Fix net.Conn type assertion

cmurphy:
Don't assert that all connections are wrapped, as they won't be if
the CloseConnOnCertChange setting is false. Only run the assertion
within a conditional for wrapped connections, where it is safe. This
prevents a panic from happening when CloseConnOnCertChange is not used.

Fixes an issue found by the rancher/rancher provisioning tests.

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -299,7 +299,7 @@ func (l *listener) getCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmurphy:
@brandond @ibuildthecloud could you review this as well?

brandond (approved):

cbron (approved):

PR #46, opened 2021-11-05
Allow for default expiration days to be loaded from env

nickgerace:
This PR enables the consumer of dynamiclistener's cert generation to specify the number of days until expiration via ENV VAR. By default, not supplying the ENV VAR will result in the _same_ behavior as before. The ENV VAR is optional and specifying will override the default. If any issues occur, we will fallback to the default value.

Partially resolves: https://github.com/rancher/rancher/issues/35366

### DIFF
--- a/factory/cert_utils.go
+++ b/factory/cert_utils.go
@@ -10,6 +10,8 @@ import (
 	"math"
 	"math/big"
 	"net"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -17,7 +19,8 @@ import (
 )
 
 const (
-	CertificateBlockType = "CERTIFICATE"
+	CertificateBlockType               = "CERTIFICATE"
+	defaultNewSignedCertExpirationDays = 365
 )
 
 func NewSelfSignedCACert(key crypto.Signer, cn string, org ...string) (*x509.Certificate, error) {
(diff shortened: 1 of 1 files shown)


nickgerace:
I cannot assign reviewers, so for visibility: this PR is awaiting review from @thedadams

factory/cert_utils.go
```diff
@@ -82,12 +85,22 @@ func NewSignedCert(signer crypto.Signer, caCert *x509.Certificate, caKey crypto.
 		return nil, err
 	}
 
+	expirationDays := defaultNewSignedCertExpirationDays
+	envExpirationDays := os.Getenv("CATTLE_NEW_SIGNED_CERT_EXPIRATION_DAYS")
+	if envExpirationDays != "" {
+		if envExpirationDaysInt, err := strconv.Atoi(envExpirationDays); err != nil {
+			logrus.Debugf("expiration days from ENV (%s) could not be converted to int (falling back to default value: %d)", envExpirationDays, defaultExpirationDays)
```
thedadams, line 92: Since this is logged only when the environment variable is set to something incorrect, I would advocate that this be `INFO` or `WARN` level.

thedadams (changes requested):

thedadams (approved):

factory/cert_utils.go
```diff
@@ -82,12 +85,22 @@ func NewSignedCert(signer crypto.Signer, caCert *x509.Certificate, caKey crypto.
 		return nil, err
 	}
 
+	expirationDays := defaultNewSignedCertExpirationDays
+	envExpirationDays := os.Getenv("CATTLE_NEW_SIGNED_CERT_EXPIRATION_DAYS")
```
kinarashah, line 89: @nickgerace The original issue ([and the comment](https://github.com/rancher/rancher/issues/35068#issuecomment-959838243))  says we're updating the expiry duration to 10 years instead of one, so where is this env var being set to 10y?

nickgerace: I will be doing that in another PR, likely to webhook itself! This is the library it uses, but good question. I should have clarified that.

kinarashah (approved):

PR #38, opened 2021-06-11
Accept IPv6 address as CN names

sjoerdsimons:
Expand the cnRegexp to also accept ipv6 addresses such as:
  * ::1
  * 2a00:1450:400e:80e::
  * 2a00:1450:400e:80e::200e

Related to: #37

Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>

PR #49, opened 2021-11-18
Fix ListenAndServe certificate expiration by preloading certs

brandond:
This change fixes an issue with how `server.ListenAndServe` loads certificates. It sets up a memory -> kubernetes -> memory Secret storage stack, which combined with lazy (on-demand) loading of certificate data, caused the certificate to not be properly renewed when it expired.

It also resolves a regression from #38 - while IPv6 addresses are no longer rejected by the CN regex, they cannot actually be stored in the CN annotation, as semicolons are not allowed in annotation keys. This also fixes handling of hostnames that are more than 41 characters long.


listener.go
```diff
@@ -405,7 +418,9 @@ func (l *listener) cacheHandler() http.Handler {
 				}
 			}
 
-			l.updateCert(h)
+			if err := l.updateCert(h); err != nil {
+				logrus.Errorf("failed to update cert with HTTP request Host header: %v", err)
+			}
```
nickgerace, line 423: Another good catch

storage/kubernetes/controller.go
```diff
@@ -89,18 +89,31 @@ func (s *storage) init(secrets v1controller.SecretController) {
...
+				Name:        s.name,
+				Namespace:   s.namespace,
+				Annotations: secret.Annotations,
+			},
+			Type: v1.SecretTypeTLS,
+			Data: secret.Data,
+		})
+		if err != nil && !errors.IsAlreadyExists(err) {
+			logrus.Warnf("Failed to create Kubernetes secret: %v", err)
+		}
```
nickgerace, line 106: Should we add retry logic here or will the warning log suffice? I'm led the believe that a non-"IsNotFound" error will likely reappear upon retry, so I'm good with this "as-is", but thought I'd ask if you considered it.

nickgerace (approved):
Thank you! The design lines up with what we discussed, so this LGTM.

storage/kubernetes/controller.go
brandond: The Kubernetes Secret also gets watched in the background, and the sync from the watch plus the write-back of any local updates to the secret will ensure that it is eventually consistent.

PR #52, opened 2021-12-03
[v0.2.3-patch] Add ability to force cert regeneration

galal-hussein:
* add ability to force cert regeneration


briandowns (approved):

brandond (approved):
LGTM, but in general if we're cherry-picking I'd like to see it come in from `git cherry-pick -x` so that the original commit ID is retained in the commit log.

PR #53, opened 2021-12-21
update config to allow for specifying experiation in days

briandowns:
Updates the config passed into the NewSignedCert function to allow for manually specifying the number of days the certificate should be set to expire from the date generated.

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -45,16 +45,15 @@ const (
 	duration365d = time.Hour * 24 * 365
 )
 
-var (
-	ErrStaticCert = errors.New("cannot renew static certificate")
-)
+var ErrStaticCert = errors.New("cannot renew static certificate")
 
-// Config contains the basic fields required for creating a certificate
+// Config contains the basic fields required for creating a certificate.
 type Config struct {
-	CommonName   string
-	Organization []string
-	AltNames     AltNames
-	Usages       []x509.ExtKeyUsage
+	CommonName       string
+	Organization     []string
+	AltNames         AltNames
+	Usages           []x509.ExtKeyUsage
+	ExpirationInDays int
 }
 
 // AltNames contains the domain names and IP addresses that will be added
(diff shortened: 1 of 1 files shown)


cert/cert.go
```diff
@@ -45,16 +45,15 @@ const (
...
 type Config struct {
-	CommonName   string
-	Organization []string
-	AltNames     AltNames
-	Usages       []x509.ExtKeyUsage
+	CommonName       string
+	Organization     []string
+	AltNames         AltNames
+	Usages           []x509.ExtKeyUsage
+	ExpirationInDays int
```
brandond, line 56: Would it be worth just making this a `*time.Duration` so that it could also go shorter than a day, if the use case called for it?

galal-hussein (approved):

cert/cert.go
briandowns: I'd be open to it if the use case ever presents itself.

brandond: I'd vote for doing so now. I know that users have asked to be able to (for example) set expiration to 10 years; if we make this a duration then we just have to add a DurationFlag to K3s/RKE2 and users can set whatever specific hours/days/years they want without having to convert to an integer number of days.

cert/cert.go
```diff
@@ -45,16 +45,15 @@ const (
...
 type Config struct {
-	CommonName   string
-	Organization []string
-	AltNames     AltNames
-	Usages       []x509.ExtKeyUsage
+	CommonName       string
+	Organization     []string
+	AltNames         AltNames
+	Usages           []x509.ExtKeyUsage
+	ExpirationInDays time.Duration
```
Oats87, line 56: I think `ExpirationInDays` might not be the correct name for this variable as it is a `time.Duration` instead of an int or something.

Oats87 (changes requested):

brandond (approved):

Oats87 (approved):

PR #54, opened 2022-01-03
Cherry-pick commits that allow setting certificate expiration date

genexpr:
Related to https://github.com/rancher/rancher/issues/35068.
This back-ports a few commits that allow to specify the webhook's certificate expiration date in days and, by default, set the certificate's expiration date to 10 years after creation.


genexpr:
When this is merged, we will cherry-pick from `master` branch a few relevant commits in the rancher-webhook itself (for branch `release/v0.1`) and update its dynamiclistener dependency version.

brandond (approved):

cbron (approved):

cmurphy (commented):
It doesn't look like the authorship of the original commits was preserved in these backported commits. `git cherry-pick` and `git cherry-pick -x` should preserve that authorship while adding you as the committer.

genexpr:
@cmurphy, you're right! I didn't know about that flag, will use it in the future. What do we do with this one though – should I redo the cherry-picking/reopen the PR, or is it OK just this once?

brandond:
Good catch. I would recommend redoing with `git cherry-pick -x` .

brandond (approved):

cmurphy (approved):

PR #57, opened 2022-04-30
Move Kubernetes Secrets storage update to goroutine

brandond:
Fixes issue where apiserver outages can block dynamiclistener from accepting new connections.

Signed-off-by: Brad Davidson <brad.davidson@rancher.com>

### DIFF
--- a/storage/kubernetes/controller.go
+++ b/storage/kubernetes/controller.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/rancher/dynamiclistener"
+	"github.com/rancher/dynamiclistener/cert"
 	"github.com/rancher/wrangler/pkg/generated/controllers/core"
 	v1controller "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/start"
@@ -13,6 +14,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 )
 
 type CoreGetter func() *core.Factory
@@ -39,10 +41,9 @@ func New(ctx context.Context, core CoreGetter, namespace, name string, backing d
 	// lazy init
 	go func() {
 		for {
-			core := core()
-			if core != nil {
-				storage.init(core.Core().V1().Secret())
-				_ = start.All(ctx, 5, core)
+			if coreFactory := core(); coreFactory != nil {
+				storage.init(coreFactory.Core().V1().Secret())
+				_ = start.All(ctx, 5, coreFactory)
 				return
 			}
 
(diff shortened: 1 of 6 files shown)


storage/kubernetes/controller.go
```diff
@@ -170,36 +171,46 @@ func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
 		}
 	}
 
+	// ensure that the merged secret actually contains data before overwriting the existing fields
+	if len(secret.Data) < 2 {
```
briandowns, line 175: Nit: Can we make "2" a constant with a descriptive name? Can be done someother time given the context of this change.

briandowns (approved):

storage/kubernetes/controller.go
brandond: Yeah, all this checking of data len isn't great. Added a more explicit check that mirrors what upstream is doing at https://github.com/kubernetes/kubernetes/blob/dcce2357ffda4b8cc132296a1e5a8d74bf9b150b/pkg/apis/core/validation/validation.go#L5577-L5583

storage/kubernetes/controller.go
```diff
@@ -152,14 +153,14 @@ func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
 	// in favor of just blindly replacing the fields on the Kubernetes secret.
 	if s.tls != nil {
 		// merge new secret with secret from backing storage, if one exists
-		if existing, err := s.storage.Get(); err == nil && existing != nil && len(existing.Data) > 0 {
+		if existing, err := s.storage.Get(); err == nil && existing != nil && isValidTLSSecret(existing) {
```
thedadams, line 156: This call to `s.storage.Get` is protected with `RLock` in the `Get` method. Since the mutex is no longer being used to protect the call to `saveInK8s`, should we use `s.Get()` here instead?

thedadams (commented):
I think one change might be needed for this.

storage/kubernetes/controller.go
brandond: Good catch. I also added a wrapper for the controller check since that's a read to the base object as well.

storage/kubernetes/controller.go
```diff
@@ -119,27 +120,31 @@ func (s *storage) init(secrets v1controller.SecretController) {
...
+	defer s.RUnlock()
+
 	existingSecret, err := s.secrets.Get(s.namespace, s.name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
 		return &v1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      s.name,
 				Namespace: s.namespace,
 			},
+			Type: v1.SecretTypeTLS,
```
thedadams, line 140: Is this a breaking change? I assume that the secrets created with this previously we `Opaque` secrets because this is missing. Which means that the `cert.IsValidTLSSecret` may fail when it isn't supposed to for upgraded instances?

thedadams (commented):
One more question. If the answer is, "It's not a breaking change" then I think it's good.

thedadams (approved):

storage/kubernetes/controller.go
brandond: I did pull the type check from the helper, just in case there were some certs stored in the `file` cache, which wouldn't set the type if the Secret wasn't ever round-tripped through Kubernetes or the merge operation.

rmweir (approved):

PR #59, opened 2022-05-11
Fix initial secret not being written to Kubernetes

brandond:
Updates to the secret that occurred before the controller was done syncing were not being written to Kubernetes. Subsequent updates to the secret would eventually get it written, but Rancher requires that the cert be written immediately. This was probably an unnecessary optimization anyway, so back it out in favor of just checking to see if the secrets controller is available.

Other fixes:
* Fix improper handling of multiple goroutines attempting to create the Kubernetes secret at the same time; this was also handled eventually but caused an unnecessary round of extra writes to the secret.
* Add warning when no cert is available on startup.
* Add warning when signing operation may change the certificate's issuer - this may indicate a problem with inconsistent CAs between nodes.
* Fix not adding IP SANs for IPv6 addresses.

Related to:
* https://github.com/rancher/rancher/pull/36932

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -290,14 +290,12 @@ func (l *listener) Accept() (net.Conn, error) {
 
 	host, _, err := net.SplitHostPort(addr.String())
 	if err != nil {
-		logrus.Errorf("dynamiclistener %s: failed to parse connection address %s: %v", l.Addr(), addr, err)
+		logrus.Errorf("dynamiclistener %s: failed to parse connection local address %s: %v", l.Addr(), addr, err)
 		return conn, nil
 	}
 
-	if !strings.Contains(host, ":") {
-		if err := l.updateCert(host); err != nil {
-			logrus.Errorf("dynamiclistener %s: failed to update cert with listener address: %v", l.Addr(), err)
-		}
+	if err := l.updateCert(host); err != nil {
+		logrus.Errorf("dynamiclistener %s: failed to update cert with connection local address: %v", l.Addr(), err)
 	}
 
 	if l.conns != nil {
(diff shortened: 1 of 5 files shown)


storage/kubernetes/controller.go
```diff
@@ -109,7 +111,9 @@ func (s *storage) init(secrets v1controller.SecretController) {
 		// local storage was empty, try to populate it
 		secret, err := s.secrets.Get(s.namespace, s.name, metav1.GetOptions{})
 		if err != nil {
-			logrus.Warnf("Failed to init Kubernetes secret: %v", err)
+			if !errors.IsNotFound(err) {
+				logrus.Warnf("Failed to init Kubernetes secret: %v", err)
+			}
 			return
```
steved, line 117: Did you still want to return if it's not found?

thedadams (approved):

listener.go
```diff
@@ -263,12 +263,18 @@ func (l *listener) Accept() (net.Conn, error) {
 	l.init.Do(func() {
 		if len(l.sans) > 0 {
 			if err := l.updateCert(l.sans...); err != nil {
-				logrus.Errorf("failed to update cert with configured SANs: %v", err)
+				logrus.Errorf("dynamiclistener %s: failed to update cert with configured SANs: %v", l.Addr(), err)
```
rmweir, line 266:
```suggestion
				logrus.Errorf("dynamiclistener %s: failed to update cert with configured SANs: %v", l.Addr().String(), err)
```

rmweir (changes requested):

listener.go
brandond: These implement the Stringer interface so explicitly calling `String()` alongside a format string is unnecessary.
https://go.dev/tour/methods/17

brandond: Glad to change it if there's a code style guideline I'm missing. There are a couple other instances of this as well here that I can change if necessary.

rmweir: Nope, I missed that mb.

rmweir (approved):
LGTM

Issue #63, opened 2022-07-25
WrapExpiration has potential busy deadloop (CPU 100%)


w13915984028:
In debugging https://github.com/harvester/harvester/issues/2558, I found it is caused by below API.

https://github.com/rancher/dynamiclistener/blob/master/listener.go#L167

```
func (l *listener) WrapExpiration(days int) net.Listener {
	ctx, cancel := context.WithCancel(context.Background())
	go func() {
		// busy-wait for certificate preload to complete
		for l.cert == nil {
			runtime.Gosched()
		}
```

Though I am not fully sure how this issue is produced, but when such condition is met, the `WrapExpiration` indeed causes deadloop.

w13915984028:
@brandond @rebeccazzzz @guangbochen

brandond:
When looking at how Rancher and K3s/rke2 I didn't see any cases where we didn't immediately bind a certificate, so the busy-wait seemed acceptable. If there are code paths that cause this to run for more than a handful of iteration then yeah we should sleep instead.

w13915984028:
@brandond 

I had a further look,  this BUG looks to be related to a couple of code change:

(1)
the `tight loop` was add in this commit, it was not in the code before, original code just wait 30 seconds and then go further.  Current loop code, in fact will block there as long as `l.cert == nil`
https://github.com/rancher/dynamiclistener/commit/284cc004e84101e4a5b9fa50c00e101095db83f6
![image](https://user-images.githubusercontent.com/31133476/181195199-006ca00b-a561-47e0-baad-6f6c09697597.png)


(2) in this commit, an additional check ` !cert.IsValidTLSSecret(secret)` was added, it will return;  and it will cause (1) goes into busy endless loop

https://github.com/rancher/dynamiclistener/commit/b1d65efb6f8309327ed7beb41463056379ec18eb#diff-52136e2f401dea328d50069c304fe222eb004d0ca7b8b12e6e6c4adbc9df6979R411
![image](https://user-images.githubusercontent.com/31133476/181195430-306dc71f-d693-43e2-9e74-35421e884f8f.png)



In my ENV,   the `secret` looked to be a `nil` value, (2) returns but the `l.cert is nil`  , and (1) is always busy looping

```
time="2022-07-25T10:07:16Z" level=warning msg="dynamiclistener [::]:8443: no cached certificate available 
for preload - deferring certificate load until storage initialization or first client request, 
id:4 name:secret:
[...]

PR #64, opened 2022-07-25
fix63 use sleep instead of force scheduling

w13915984028:
fix https://github.com/rancher/dynamiclistener/issues/63

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -7,7 +7,6 @@ import (
 	"crypto/x509"
 	"net"
 	"net/http"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


brandond (approved):

listener.go
```diff
@@ -165,7 +164,7 @@ func (l *listener) WrapExpiration(days int) net.Listener {
 	go func() {
 		// busy-wait for certificate preload to complete
```
brandond, line 165: Nit: correct the comment as well
```suggestion
		// loop on short sleeps until certificate preload completes
```

w13915984028: updated, thanks.

brandond:
LGTM; what do you see for CPU load with this change? I'm curious how long the listener remains operating without a certificate in this use case, and whether polling at 1000hz is still runs the CPU more than we'd like.

w13915984028:
Before, 1 cpu core 100%

Now:
top inside POD
image

prometheus monitoring:
image


In each second, `harvester` POD has 10+ API server driven actions.
Simply sleep 1000 times per second won't use too much CPU resources.

rmweir (approved):

cbron (approved):

PR #65, opened 2022-07-28
Removes wait loop for listener certs.

KevinJoiner:


listener.go
```diff
@@ -439,6 +439,7 @@ func (l *listener) loadCert(currentConn net.Conn) (*tls.Certificate, error) {
 	}
 
 	l.cert = &cert
+	close(l.certIsSetSig)
```
rmweir, line 442: This feels it may be an improper way to use a channel. We are never actually passing anything into it so what we really want is probably just a mutex.

rmweir (changes requested):

listener.go
KevinJoiner: We are sending information that the cert has been populated across routines which I believe is the correct way to use channels. We could send an empty struct on the channel, but closing the channel allows everyone waiting and successive calls to be unblocked. 
IMO a mutex is better suited for stopping multiple threads/routines from accessing the same data at the same time, which is not really what we are trying to accomplish. We are more interested in knowing when the cert is ready. I believe this use case is similar to how channels are used in [time.After](https://pkg.go.dev/time#After). 
All that being said, we can accomplish our goal with either implementation, and neither is inherently better than the other.

cbron: I think using channels here makes sense, however I would much rather merge this for 2.6.7: https://github.com/rancher/dynamiclistener/pull/64
And then come back to this at the beginning of 2.6.8. We know that sleeping is a valid solution. If this PR misses a spot that might load the cert, and then that fails to close the channel we could easily introduce bugs. But in 2.6.8 I'm fine with this solution.

w13915984028: With a channel, do you need to use `select` to add process of `ctx.Done()` ?

In my ENV, here it never meet `<-l.certIsSetSig` due to `if !cert.IsValidTLSSecret(secret) ` in `loadCert`,  and it looks necessary to add `error/warn log` before `if !cert.IsValidTLSSecret(secret) ` returns.

```
			select {
			case <-ctx.Done():
				return
                        case <-l.certIsSetSig: ...
```

brandond: We do this (block on reading from a channel that is only closed, never written to) all over the place in other rancher projects. It's a solid pattern.

listener.go
```diff
@@ -58,11 +57,12 @@ func NewListener(l net.Listener, storage TLSStorage, caCert *x509.Certificate, c
 			FilterCN:            allowDefaultSANs(config.SANs, config.FilterCN),
 			ExpirationDaysCheck: config.ExpirationDaysCheck,
 		},
-		Listener:  l,
-		storage:   &nonNil{storage: storage},
-		sans:      config.SANs,
-		maxSANs:   config.MaxSANs,
-		tlsConfig: config.TLSConfig,
+		Listener:     l,
+		certIsSetSig: make(chan struct{}),
```
brandond, line 61: Nit: we tend to call these ready channels in other projects, what about:
```suggestion
		certReady: make(chan struct{}),
```

listener.go
w13915984028: but you never meet deadloop before,  now the deadloop comes

listener.go
```diff
@@ -150,23 +150,23 @@ type listener struct {
...
 func (l *listener) WrapExpiration(days int) net.Listener {
 	ctx, cancel := context.WithCancel(context.Background())
 	go func() {
-		// busy-wait for certificate preload to complete
-		for l.cert == nil {
-			runtime.Gosched()
-		}
+
+		// wait for cert to be set, this will unblock when the channel is closed
+		<-l.certIsSetSig
```
brandond, line 169: Can we confirm that the channel is wired up and will be closed on all code paths that reach this wait?

w13915984028: In `if !cert.IsValidTLSSecret(secret) in loadCert`,  it won't.

listener.go
brandond: Yeah I suspect that the channel is not getting closed on all code paths, which means that either:
* there are other paths to setting `l.cert` that aren't closing the channel
* the channel in the struct is being set too late, and is changed after the goroutine blocks on the read

listener.go
KevinJoiner: > Can we confirm that the channel is wired up and will be closed on all code paths that reach this read?

The channel is created in the `NewListner` function, the only location where a `listener` struct is instantiated.
> In if !cert.IsValidTLSSecret(secret) in loadCert, it won't.

In `if !cert.IsValidTLSSecret(secret)` in `loadCert` it won't be closed because `l.cert` will still be nil. Which I believe is the desired behavior and the same behavior seen with the previous implementation. Since`l.cert` is private it can only be set at the end of `loadCert`.

listener.go
KevinJoiner: @w13915984028 I like the suggestion of adding the context in the select so we can kill the routine with a cancel.
Before this change, instead of a deadlock we had a spinlock, which behaves the same, except the spin lock uses CPU time. @brandond, could you double-check me? But I believe we can guarantee both those scenarios you suggested do not occur because we are dealing with a private struct and member variable.

listener.go
w13915984028: Is it possible that the `cert` may be load multi times ? 

  if true, then the `WrapExpiration` is not very reasonable, it seems should be triggered right after each successful `loadCert`, as you should run `WrapExpiration` from the beginning, it check first then decide long or short timer;  but the current code, the `WrapExpiration` will still be in laster timer

  if not, then it's still reasonable to trigger `WrapExpiration` in `loadCert`, only after a true load happens, you dont need a channel

whenever a `cert` is scucessfully loaded, stop the potential previous `WrapExpiration`, and start a new `WrapExpiration`

KevinJoiner: `l.cert` can be updated after it is initially set by setting the version to empty string `l.version = ""`, as seen [here](https://github.com/rancher/dynamiclistener/blob/master/listener.go#L384-L387). Updating the cert would not require us to start a new `WrapExperation`.
> whenever a cert is scucessfully loaded, stop the potential previous WrapExpiration, and start a new WrapExpiration

There are some reasons I would not want to do this.
1. We can not stop `WrapExperation` unless we close the listener, which we do not want to do.
2. `WrapExperation` is public, so we would have to keep track of all calls to the function to perform proper clean up of all go routines. 
3.  I don't think we want to potentially break other projects by changing the function signature of `WrapExperation` to work around 1 & 2
4. By not starting a new instance of `WrapExperation`  we have the chance of updating cert expiration before 6 hours which is an acceptable outcome.

brandond (approved):

w13915984028 (approved):
LGTM, thanks.

rmweir (approved):

brandond:
Anyone up to merge this? I can do so myself if no-one else is available.

KevinJoiner:
@cbron would like to wait for 2.7.0 to add this and get an issue with QA test for both Rancher and RKE2.

KevinJoiner:
Related to https://github.com/rancher/dynamiclistener/issues/63

yaocw2020:
https://github.com/rancher/dynamiclistener/blob/2b62d5cc694d566dd8f3f67eb7b0f6bb46266a65/listener.go#L417
It looks very strange.

yaocw2020:
Related to https://github.com/harvester/harvester/issues/4001

brandond:
This still looks good to me, and doesn't appear to have any conflicts that would prevent it from being merged. @cbron do you have any remaining concerns?

brandond:
Merging as per Slack conversation

PR #66, opened 2022-10-11
Marking additional connections as ready

MbolotSuse:
Most connections were not marked as ready despite having retrieved a valid cert. This change makes all connections which successfully retrieved a cert get marked as ready

Related to rancher/rancher#39094


MbolotSuse:
Still working on some tests, but would appreciate any comments on the approach/functionality as I put together some tests for `getCertificate`

brandond (approved):

MbolotSuse:
## Overview

This pr is related to rancher/rancher#39094. Basically, not every connection which got a valid cert was marked as ready when a valid cert was retrieved for it. This lead to issues where the UI maintained some connections with an old cert event though a newer cert was being used for new connections. Since the old cert was still being used, this prevented user's Browsers from recognizing that it needed to re-prompt the user to accept the new self-signed cert, leading to constant websocket disconnect/reconnect attempts. For more information on this issue, see [this comment](https://github.com/rancher/rancher/issues/39094#issuecomment-1275107399). 

This PR aims to fix this by marking any connection which we retrieved a valid cert for as ready. Since `loadCert` is only called with an active connection in one place, we only need to perform this check in one side function (`getCertificate`). We also only attempt to mark a connection as ready if we have an active connection (to avoid nil pointer exceptions) and if `l.conns` has been initializing (indicating that consuming function requested that we close connections on cert change - see [here](https://github.com/rancher/dynamiclistener/blob/401fafb7e6dfec2b55aa0008f51640e8a8bb3fce/listener.go#L71-L76) for more). 

## Changes
- Added a check to `getCertificate` which marks a connection as ready if:
  - We successfully retrieved a cert for this connection (cert is not nil and err is nil)
  - This listener was configured to close connections on cert change (`l.conns` is not nil)
  - We have a current connection
- Removes the logic in `loadCertificate` which marks a connection which forced a cert change as ready. This is not done by `getCertificate`, and would be redundant in the new structure.

MbolotSuse:
Got some tests added, and also pushed a commit which bumps the go version for this to 1.19.

listener.go
```diff
@@ -345,8 +345,20 @@ func (l *listener) getCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate,
...
 	}
-
-	return l.loadCert(newConn)
+	connCert, err := l.loadCert(newConn)
+	if connCert != nil && err == nil && newConn != nil && l.conns != nil {
+		// if we were successfully able to load a cert and are closing connections on cert changes, mark newConn ready
+		// this will allow us to close the connection if a future connection forces the cert to re-load
+		wrapper, ok := newConn.(*closeWrapper)
+		if !ok {
+			logrus.Infof("will not mark non-close wrapper connection from %s to %s as ready", newConn.RemoteAddr(), newConn.LocalAddr())
```
cmurphy, line 354: This could get noisy, I might downgrade this to Debugf

const_test.go
```diff
@@ -0,0 +1,100 @@
+package dynamiclistener
```
cmurphy, line 1: nit: the file name strikes me as weird, there are no tests so the `_test.go` suffix is inappropriate and there are technically no consts. Maybe `test_vars.go` or `test_data.go`?

cmurphy (approved):
Nice job on the tests :+1:

listener.go
brandond: This looks kinda like development logging that could just go away entirely. The connection will not be wrapped by closeWrapper in most non-Rancher projects where we don't ask for the close-on-change behavior.

const_test.go
brandond: I agree. Maybe move it into a tests directory, and import it in `listener_test.go`?

brandond (changes requested):

listener.go
MbolotSuse: I'll downgrade to debug - I think that's more consistent with project behavior. 

> This looks kinda like development logging that could just go away entirely. The connection will not be wrapped by closeWrapper in most non-Rancher projects where we don't ask for the close-on-change behavior.

@brandond In those cases, I would expect that `l.conns == nil` and this code to never be invoked. I want this debug to be in place because the population of l.conns indicates that you wanted the close-on-change behavior, but we weren't able to provide it because of improper wrapping. See here: https://github.com/rancher/dynamiclistener/blob/401fafb7e6dfec2b55aa0008f51640e8a8bb3fce/listener.go#L71-L76 for what I mean on expecting this block to only be called when you asked for close-on-change behavior.

brandond: In that case would it be an error? Informational says it's something that a user would want to know about. If it indicates a problem in our code, lets make it an error. If it doesn't matter to the user lets make it a debug or trace.

MbolotSuse: From a strictly technical perspective this could probably be best classified as an error. But I think @cmurphy 's concern still stands in that case - if we are wrong this will flood people's logs.

const_test.go
KevinJoiner: @cmurphy `_test.go`'s intended use is to notify the compiler that a file should only be used when compiling a test. Since these variables are not used outside of tests, I wouldn't want to compile them in all binaries that import dynamiclistener. 

@brandond's suggestion would stop us from compiling in these keys, but I think a test directory without any tests is more confusing than a `vars_test.go` file without any tests.

I am biased because I have seen/used this design pattern for holding large string data that is used only for tests.
I think the alternative is just to paste the variable at the end of the test they are used in, so the file is still readable.

KevinJoiner (approved):

const_test.go
brandond: I haven't poked at the certs you're embedding here - how long are they valid for? Are we in danger of the automatic renewal code breaking the test when they are within 90 days of expiring?

MbolotSuse: > I haven't poked at the certs you're embedding here - how long are they valid for?

They were valid for 100 years. 

> This library has a ton of cert creation code, Is there any reason we can't just generate the certs as part of the test

I had a long comment here about additional test complexity and not wanting one test for one part of this code to depend on another part, when I decided to just try it to see if I could put something together.

Turns out this worked out pretty well. The test now generates it's own certs instead of using the const. I've eliminated the `const_test` file as a consequence. Thanks @brandond for the suggestion.

listener.go
brandond: Fair enough. Are you still planning on downgrading this to a debug? LGTM otherwise.

MbolotSuse: Yep - I'll downgrade to debug and push this up. Sorry for the delay.

MbolotSuse: Resolving, this now uses `Debugf`

cmurphy (approved):

brandond (approved):

PR #78, opened 2023-07-07
Allow multiple (intermediate) CA certs

knoppiks:
This PR tries to enable the dynamiclistener to hand out not only the first signing certificate it finds in the CA chain.

I stumbled on [this discussion](https://github.com/k3s-io/k3s/discussions/7596) while searching for the exact problem in k3s, where @brandond pointed to this repository.

Maybe it would be good to add tests but being not a go expert I struggle to comprehend the tests conducted on the listener.

### DIFF
--- a/factory/ca.go
+++ b/factory/ca.go
@@ -28,6 +28,9 @@ func GenCA() (*x509.Certificate, crypto.Signer, error) {
 // Deprecated: Use LoadOrGenCAChain instead as it supports intermediate CAs
 func LoadOrGenCA() (*x509.Certificate, crypto.Signer, error) {
 	chain, signer, err := LoadOrGenCAChain()
+	if err != nil {
+		return nil, nil, err
+	}
 	return chain[0], signer, err
 }
 
@@ -69,7 +72,10 @@ func loadCA() ([]*x509.Certificate, crypto.Signer, error) {
 
 func LoadCA(caPem, caKey []byte) (*x509.Certificate, crypto.Signer, error) {
 	chain, signer, err := LoadCAChain(caPem, caKey)
-	return chain[0], signer, err
+	if err != nil {
+		return nil, nil, err
+	}
+	return chain[0], signer, nil
 }
 
 func LoadCAChain(caPem, caKey []byte) ([]*x509.Certificate, crypto.Signer, error) {
@@ -93,6 +99,9 @@ func LoadCAChain(caPem, caKey []byte) ([]*x509.Certificate, crypto.Signer, error
 // Deprecated: Use LoadCertsChain instead as it supports intermediate CAs
 func LoadCerts(certFile, keyFile string) (*x509.Certificate, crypto.Signer, error) {
 	chain, signer, err := LoadCertsChain(certFile, keyFile)
+	if err != nil {
+		return nil, nil, err
+	}
 	return chain[0], signer, err
 }
 
--- a/factory/ca.go
+++ b/factory/ca.go
@@ -25,18 +25,25 @@ func GenCA() (*x509.Certificate, crypto.Signer, error) {
 	return caCert, caKey, nil
 }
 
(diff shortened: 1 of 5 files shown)


brandond:
This looks good structurally, however I don't think we can change the call signatures of existing exported functions in order to avoid breaking users of this library. We'd probably need to modify this PR to include new functions that return `[]x509.Certificate`, and having those be used to load the cert, instead of simply changing the existing ones.

knoppiks:
@brandond Please have a look. I tried my best with the naming, maybe you have better suggestions.

factory/ca.go
```diff
@@ -53,14 +60,19 @@ func LoadOrGenCA() (*x509.Certificate, crypto.Signer, error) {
...
 
-func loadCA() (*x509.Certificate, crypto.Signer, error) {
-	return LoadCerts("./certs/ca.pem", "./certs/ca.key")
+func loadCA() ([]*x509.Certificate, crypto.Signer, error) {
+	return LoadCertsChain("./certs/ca.pem", "./certs/ca.key")
 }
 
 func LoadCA(caPem, caKey []byte) (*x509.Certificate, crypto.Signer, error) {
+	chain, signer, err := LoadCAChain(caPem, caKey)
+	return chain[0], signer, err
```
brandond, line 72: You'll get a panic here if `LoadCAChain` returns a non-nil error, as you can't do `chain[0]` on an empty array.
```suggestion
	if err != nil {
		return nil, nil, err
	}
	return chain[0], signer, nil
```

factory/ca.go
```diff
@@ -70,15 +82,21 @@ func LoadCA(caPem, caKey []byte) (*x509.Certificate, crypto.Signer, error) {
...
 	}
 
-	return cert, signer, nil
+	return certs, signer, nil
 }
 
+// Deprecated: Use LoadCertsChain instead as it supports intermediate CAs
 func LoadCerts(certFile, keyFile string) (*x509.Certificate, crypto.Signer, error) {
+	chain, signer, err := LoadCertsChain(certFile, keyFile)
+	return chain[0], signer, err
```
brandond, line 96: same issue as above with `chain[0]`

listener.go
```diff
@@ -34,7 +34,12 @@ type SetFactory interface {
 	SetFactory(tls TLSFactory)
 }
 
+// Deprecated: Use NewListener2 instead as it supports intermediate CAs
 func NewListener(l net.Listener, storage TLSStorage, caCert *x509.Certificate, caKey crypto.Signer, config Config) (net.Listener, http.Handler, error) {
+	return NewListener2(l, storage, []*x509.Certificate{caCert}, caKey, config)
+}
+
+func NewListener2(l net.Listener, storage TLSStorage, caCert []*x509.Certificate, caKey crypto.Signer, config Config) (net.Listener, http.Handler, error) {
```
brandond, line 42: Give this a more descriptive name

```suggestion
func NewListenerWithChain(l net.Listener, storage TLSStorage, caCert []*x509.Certificate, caKey crypto.Signer, config Config) (net.Listener, http.Handler, error) {
```

brandond (changes requested):
couple nits on error handing and naming, looks good otherwise!

factory/ca.go
knoppiks: Very good hint, thank you. Found two more places with the same problem.

brandond (approved):

knoppiks:
Any news here? I'd really like to tackle the k3s portion of this fix.

caroline-suse-rancher:
Hey @knoppiks we're in code freeze at the moment for July releases, but after that we'll be looking for one more reviewer/approver for this PR. I will bring it up with the team after freeze is lifted!

brandond:
We can get it merged here, and then hold off on updating anything in K3s until after the freeze is over. This repo is not subject to code freeze.

briandowns (approved):

dereknola (approved):

PR #80, opened 2023-08-22
updated deps for k8s 1.27

vardhaman22:
related issue: https://github.com/rancher/rancher/issues/41840

with the k8s deps upgrade this also fixes server.go **wrong listener function name**. With this commit https://github.com/rancher/dynamiclistener/pull/78/commits/8f13b193a1028827453c064efa89629725b6cdc7 the listener function was renamed (ref PR: https://github.com/rancher/dynamiclistener/pull/78).. but it was left to be renamed in server.go


kinarashah:
changes LGTM, will revisit for merging after we update replace section

kinarashah (approved):

PR #81, opened 2023-09-21
wait for storage init when saving secret

vardhaman22:
## Issue
issue: https://github.com/rancher/rancher/issues/42961
after the async changes in storage code there is a race condition in code for the **serving-cert** creation for listener. 

In rancher when [listenAndServe](https://github.com/rancher/rancher/blob/d101c27fe5eb6bfd1992165b2f08c9fc02a2a55f/pkg/tls/tls.go#L88) is called, it ultimately tries to create serving-cert secret. But if the storage is not marked initialized when it tries to save that secret the code just returns from [there](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L164) and the secret is never saved.(happening in most cases)

## Solution
added a go routine to save the secret in k8s when storage is initialized.

Tested this solution in rancher. CI is passing. [PR](https://github.com/rancher/rancher/pull/42891) [TAG](https://github.com/vardhaman22/dynamiclistener/commits/v1.27.5-fix)
 
## Details:
function calls stack ->

server.ListenAndServe ([next](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/server/server.go#L78) ) => [listener.Accept](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/listener.go#L271) ([next](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/listener.go#L274)) => [l.updateCert ](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/listener.go#L374) ([next](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/listener.go#L403))=> [l.storage.Update](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/listener.go#L509) ([next](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/listener.go#L513))

ultimately the storage Update function tries to [save](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L232) the secret.
During this save process it [checks](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L163) if the storage init is complete. The storage init condition was [changed](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L249) in [commit](https://github.com/rancher/dynamiclistener/commit/1c60bf414fd2331505a8dbd039e527ee3c7a94dc)). 
Since the storage init is now made asynchronous, there is now a race condition i.e. initialized is being set to true after the code returns from [here](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L164), so the secret is never saved which is causing issues in rancher CI.

### DIFF
--- a/storage/kubernetes/controller.go
+++ b/storage/kubernetes/controller.go
@@ -26,6 +26,7 @@ func Load(ctx context.Context, secrets v1controller.SecretController, namespace,
 		namespace: namespace,
 		storage:   backing,
 		ctx:       ctx,
+		initSync:  &sync.Once{},
 	}
 	storage.init(secrets)
 	return storage
@@ -37,6 +38,7 @@ func New(ctx context.Context, core CoreGetter, namespace, name string, backing d
 		namespace: namespace,
 		storage:   backing,
 		ctx:       ctx,
+		initSync:  &sync.Once{},
 	}
 
 	// lazy init
@@ -62,6 +64,7 @@ type storage struct {
 	ctx             context.Context
 	tls             dynamiclistener.TLSFactory
 	initialized     bool
+	initSync        *sync.Once
 }
 
 func (s *storage) SetFactory(tls dynamiclistener.TLSFactory) {
@@ -161,6 +164,20 @@ func (s *storage) targetSecret() (*v1.Secret, error) {
 
 func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
 	if !s.initComplete() {
+		// Start a goroutine to attempt to save the secret later, once init is complete.
+		// If this was already handled by initComplete, it should be a no-op, or at worst get
+		// merged with the Kubernetes secret.
+		go s.initSync.Do(func() {
+			if err := wait.Poll(100*time.Millisecond, 15*time.Minute, func() (bool, error) {
+				if !s.initComplete() {
+					return false, nil
+				}
+				_, err := s.saveInK8s(secret)
+				return true, err
+			}); err != nil {
+				logrus.Errorf("Failed to save TLS secret after controller init: %v", err)
+			}
+		})
 		return secret, nil
 	}
 


brandond:
This is already supposed to be handled by the bit at
https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L91-L96

You are correct that the write to the Kubernetes Secret is skipped if the Secret controller hasn't initialized yet, but the `syncStorage` call should handle that gap by syncing the backing storage into Kubernetes as soon as controller initialization completes, without having to poll.

If `saveInk8s` returns early because the controller has not yet initialized, the secret still gets written to the backing storage for later sync, here:
https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L232-L243

vardhaman22:
> This is already supposed to be handled by the bit at
> 
> https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L91-L96
> 
> You are correct that the write to the Kubernetes Secret is skipped if the Secret controller hasn't initialized yet, but the `syncStorage` call should handle that gap by syncing the backing storage into Kubernetes as soon as controller initialization completes, without having to poll.
> 
> If `saveInk8s` returns early because the controller has not yet initialized, the secret still gets written to the backing storage for later sync, here:
> 
> https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L232-L243
> 
> K3s and RKE2 use `kubernetes` storage, backed by `file`, and this seems to work correctly. Is this perhaps an issue that is unique to use with the `static` or `memory` storages?

actually i am not actually sure what it can be .. was inspecting the logs of rancher on 2.8 head and the ones with CI failures.
So in the CI failures there is no log entry like
```
creating new TLS Creating new TLS secret for cattle-system/serving-cert
```
so i tried to debug that and that was being skipped at the line where i added poll.

vardhaman22:
> This is already supposed to be handled by the bit at
> 
> https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L91-L96
> 
> You are correct that the write to the Kubernetes Secret is skipped if the Secret controller hasn't initialized yet, but the `syncStorage` call should handle that gap by syncing the backing storage into Kubernetes as soon as controller initialization completes, without having to poll.
> 
> If `saveInk8s` returns early because the controller has not yet initialized, the secret still gets written to the backing storage for later sync, here:
> 
> https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L232-L243
> 
> K3s and RKE2 use `kubernetes` storage, backed by `file`, and this seems to work correctly. Is this perhaps an issue that is unique to use with the `static` or `memory` storages?

did a little more digging and it seems the memory storage is being used by rancher and the code which you mentioned is leading to the following lines:
https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/memory/memory.go#L35-L45

but i think the secret is not getting stored in kubernetes. 

One more thing if we use the current master of rancher/dynamiclistener in rancher and run it locally.... the secret gets generated when we first go to https://localhost:8443 .. i.e. as soon as i load that page this log pops up
```
2023/09/22 00:38:26 [INFO] Creating new TLS secret for cattle-system/serving-cert (count: 5): map[listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/cn-192.168.18.37:192.168.18.37 listener.cattle.io/cn-__1-f16284:::1 listener.cattle.io/cn-localhost:localhost listener.cattle.io/cn-rancher.cattle-system:rancher.cattle-system listener.cattle.io/fingerprint:SHA1=8E1162EFF5378965944FECDC6ED5EB713D6D46B0]
```

maybe CI is failing because it does not hit a client request .. but not totally sure.

storage/kubernetes/controller.go
```diff
@@ -160,7 +160,16 @@ func (s *storage) targetSecret() (*v1.Secret, error) {
 }
 
 func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
-	if !s.initComplete() {
+
+	// wait for storage init
+	err := wait.PollUntilContextTimeout(context.Background(), 100*time.Millisecond, 1*time.Minute, true, func(ctx context.Context) (done bool, err error) {
```
brandond, line 165: 1. Polling with a context, but then using `context.Background()` doesn't make a lot of sense
2. This makes the `saveInk8s` function block for up to a minute when updating secrets if the apiserver isn't available, which is unacceptable for K3s. dynamiclistener is started very early, and the apiserver later wired into its request handlers. Dynamiclistener needs to fail fast and write the secret through to the backing storage immediately, if the secret controller isn't up yet.

brandond (changes requested):

storage/kubernetes/controller.go
brandond: Issues with use of context aside, It sounds like we maybe need to make this configurable so that different projects can have their desired behavior. 

If the apiserver is unavalable:
* K3s wants the backing storage updated immediately, and the Kubernetes secret synced later once the apiserver is up.
* Rancher wants to block on the apiserver coming up, and not update the backing storage until after the secret has been updated.

felipe-colussi:
@brandond  What seems to be happening  on rancher is that the `secret, err := s.Get()` is returning nill.
That locks seems to be called before the memory storage  being populated.  

With that we are not able to do the sync. 

ON this [code](https://github.com/rancher/dynamiclistener/blob/0132d96ec2c50a73e21d6ea2d19711bd268e6a45/storage/kubernetes/controller.go#L117C1-L126C3): 
```
                // local storage was empty, try to populate it
                secret, err = s.secrets.Get(s.namespace, s.name, metav1.GetOptions{})
                if err != nil {
                        if !errors.IsNotFound(err) {
                                logrus.Warnf("Failed to init Kubernetes secret: %v", err)
                        }
                } else {
                        updateStorage = true
                }
        }
```
I was thinking of adding a else   like:  

```

                        if !errors.IsNotFound(err) {
                                logrus.Warnf("Failed to init Kubernetes secret: %v", err)
                        } else {
                               retry Up to X times; 
                        } 


```

DO you think that it would generate a problem  for K3s or RKE2?

brandond:
OK, so the race condition is between the `syncStorage` function seeing an empty secret in the backing storage, and the `saveInK8s` function returning early because init is not complete, followed by `update` to writing to the backing storage.

In this case, `initComplete` sees that local storage is empty and tries to populate it from the Kubernetes secret, which may overwrite what was just put there by `update`. At the same time, the original secret from the backing store is never written to the Kubernetes secret, because we skipped both the code in `saveInK8s`, and the code in `initComplete`, that would handle writing it there.

I don't think we can just fix this by changing the lock scope on `initComplete`, as we need it to do work without holding a lock in case the kubernetes calls hang. 

What if we did something like this:

```golang
type storage struct {
	sync.RWMutex

	namespace, name string
	storage         dynamiclistener.TLSStorage
	secrets         v1controller.SecretController
	ctx             context.Context
	tls             dynamiclistener.TLSFactory
	initialized     bool
	initSync *sync.Once // add this
}
```

```golang
	storage := &storage{
		name:      name,
		namespace: namespace,
		storage:   backing,
		ctx:       ctx,
		initSync: *sync.Once{}, // add this
	}
```

```golang
func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
	if !s.initComplete() {
		// Start a goroutine to attempt to save the secret later, once init is complete.
		// If this was already handled by initComplete, it should be a no-op, or at worst get
		// merged with the Kubernetes secret.
		go s.initSync.Do(func(){
			for !s.initComplete() {
				time.Sleep(100*time.Millisecond)
			}
			s.saveInK8s(secret)
		})
		return secret, nil
	}
```

felipe-colussi:
I do not have much domain over this code.

But this solution probably can work. I'll try it tomorrow.

For what I'm seeing at the beginning the TLS (MEMORY) and secrets (k8s) are being created without a name nor namespace.

The following logs are from a code that I ran with a sleep as mentioned on my [previous comment].(https://github.com/rancher/dynamiclistener/pull/81#issuecomment-1730211875)
```
2023/09/21 20:40:49 [INFO] Active TLS secret / (ver=) (count 4): map[listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/cn-localhost:localhost listener.cattle.io/cn-rancher.cattle-system:rancher.cattle-system listener.cattle.io/fingerprint:SHA1=1BF78EB3F57CD38EFE434C3B94A2FEDDB8D14795]
2023/09/21 20:40:49 [INFO] Active TLS secret / (ver=) (count 1): map[listener.cattle.io/cn-10.43.57.17:10.43.57.17 listener.cattle.io/fingerprint:SHA1=045CBA93932C7C8D0A4EB59AFF921D2136D8F546]
2023/09/21 20:40:50 [ERROR] FELIPE - Retry
2023/09/21 20:40:50 [INFO] Active TLS secret cattle-system/serving-cert (ver=2127) (count 4): map[listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/cn-localhost:localhost listener.cattle.io/cn-rancher.cattle-system:rancher.cattle-system listener.cattle.io/fingerprint:SHA1=1BF78EB3F57CD38EFE434C3B94A2FEDDB8D14795]
2023/09/21 20:40:50 [ERROR] Creating secret cattle-system / serving-cert
2023/09/21 20:40:50 [ERROR] FELIPE - Retry
2023/09/21 20:40:50 [ERROR] Creating secret cattle-system / tls-rancher-internal

2023/09/21 20:40:50 [INFO] Active TLS secret cattle-system/tls-rancher-internal (ver=2131) (count 1): map[listener.cattle.io/cn-10.43.57.17:10.43.57.17 listener.cattle.io/fingerprint:SHA1=045CBA93932C7C8D0A4EB59AFF921D2136D8F546]
2023/09/21 20:41:43 [INFO] Active TLS secret cattle-system/tls-rancher-internal (ver=3177) (count 2): map[listener.cattle.io/cn-10.43.57.17:10.43.57.17 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/fingerprint:SHA1=3800A67D5B8C6CF7D8F2A0DC6670BB9DE9EF3E36]
2023/09/21 20:41:43 [INFO] Active TLS secret cattle-system/tls-rancher-internal (ver=3180) (count 2): map[listener.cattle.io/cn-10.43.57.17:10.43.57.17 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/fingerprint:SHA1=B3F1570160555C4CEBA5D6DAA6B915448CE32024]
2023/09/21 20:41:43 [INFO] Active TLS secret cattle-system/tls-rancher-internal (ver=3196) (count 2): map[listener.cattle.io/cn-10.43.57.17:10.43.57.17 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/fingerprint:SHA1=3800A67D5B8C6CF7D8F2A0DC6670BB9DE9EF3E36]
```

On previous runs (without any sleep / retry) the listener was activating (creating on memory) a secret for the "serving-cert" without name nor namespace and wasn't able to reconcile it.

Thats why I thought that the retry on the syncStorage would be a good idea.

```
2023/09/19 16:11:55 [INFO] Active TLS secret / (ver=) (count 4): map[listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/cn-localhost:localhost listener.cattle.io/cn-rancher.cattle-system:rancher.cattle-system listener.cattle.io/fingerprint:SHA1=CFD32429A9839CC2A109108F9B1F0388C9310F65]
2023/09/19 16:11:55 [INFO] Creating new TLS secret for cattle-system/tls-rancher-internal (count: 1): map[listener.cattle.io/cn-10.43.178.139:10.43.178.139 listener.cattle.io/fingerprint:SHA1=C1C5096EEF701383B08EA5C4268C092AD2491671]
2023/09/19 16:11:55 [INFO] Active TLS secret cattle-system/tls-rancher-internal (ver=1799) (count 1): map[listener.cattle.io/cn-10.43.178.139:10.43.178.139 listener.cattle.io/fingerprint:SHA1=C1C5096EEF701383B08EA5C4268C092AD2491671]
2023/09/19 16:11:55 [INFO] Updating TLS secret for cattle-system/tls-rancher-internal (count: 1): map[listener.cattle.io/cn-10.43.178.139:10.43.178.139 listener.cattle.io/fingerprint:SHA1=C1C5096EEF701383B08EA5C4268C092AD2491671]
2023/09/19 16:13:56 [INFO] Updating TLS secret for cattle-system/tls-rancher-internal (count: 2): map[listener.cattle.io/cn-10.43.178.139:10.43.178.139 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/fingerprint:SHA1=2EBDEE13B5A9F8AF5A8AD1076060936CAAF2701F]
2023/09/19 16:13:56 [INFO] Active TLS secret cattle-system/tls-rancher-internal (ver=3486) (count 2): map[listener.cattle.io/cn-10.43.178.139:10.43.178.139 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/fingerprint:SHA1=2EBDEE13B5A9F8AF5A8AD1076060936CAAF2701F]
2023/09/19 16:13:56 [INFO] Updating TLS secret for cattle-system/tls-rancher-internal (count: 2): map[listener.cattle.io/cn-10.43.178.139:10.43.178.139 listener.cattle.io/cn-172.17.0.3:172.17.0.3 listener.cattle.io/fingerprint:SHA1=2EBDEE13B5A9F8AF5A8AD1076060936CAAF2701F]
```

brandond:
yeah, if they have no name or namespace, they haven't been sync'd into kubernetes yet - those fields remain empty until it's been through kubernetes a least once.

vardhaman22:
> OK, so the race condition is between the syncStorage function seeing an empty secret in the backing storage, and the saveInK8s function returning early because init is not complete, followed by update to writing to the backing storage.
> 
> In this case, initComplete sees that local storage is empty and tries to populate it from the Kubernetes secret, which may overwrite what was just put there by update. At the same time, the original secret from the backing store is never written to the Kubernetes secret, because we skipped both the code in saveInK8s, and the code in initComplete, that would handle writing it there.
> 
> I don't think we can just fix this by changing the lock scope on initComplete, as we need it to do work without holding a lock in case the kubernetes calls hang.
> 
> What if we did something like this:
> 
> ```
> type storage struct {
> 	sync.RWMutex
> 
> 	namespace, name string
> 	storage         dynamiclistener.TLSStorage
> 	secrets         v1controller.SecretController
> 	ctx             context.Context
> 	tls             dynamiclistener.TLSFactory
> 	initialized     bool
> 	initSync *sync.Once // add this
> }
> ```
> 
> ```
> 
> 	storage := &storage{
> 		name:      name,
> 		namespace: namespace,
> 		storage:   backing,
> 		ctx:       ctx,
> 		initSync: &sync.Once{}, // add this
> 		initSync: *sync.Once{}, // add this
> 	}
> ```
> 
> ```
> func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
> 	if !s.initComplete() {
> 		// Start a goroutine to attempt to save the secret later, once init is complete.
> 		// If this was already handled by initComplete, it should be a no-op, or at worst get
> 		// merged with the Kubernetes secret.
> 		go s.initSync.Do(func(){
> 			for !s.initComplete() {
> 				time.Sleep(100*time.Millisecond)
> 			}
> 			s.saveInK8s(secret)
> 		})
> 		return secret, nil
> 	}
> ```

i think this should work...updated the PR with the suggestions. It is replacing the poll with a go-routine which will unblock the process to make it work for RKE2 and K3s.

brandond (commented):
I'm not going to approve this myself, since I suggested it - but I have spun up a PR in k3s to test it over there to confirm it doesn't regress anything.

* https://github.com/k3s-io/k3s/pull/8401

felipe-colussi:
@brandond, @bfbachmann   Can you confirm to me if this make sense? 

I'm not able to test that on rancher due to some other CI fails. 

That said I was taking a deeper look into the code and I don't get why we are using the `s.syncStorage` inside the `storage.init` function.

At a first glance that function seems to be used to sync the memory storage with the k8s one, that made sense wile it was blocking and being executed synchronously. That way during the execution all would be ready, IMO once it was moved to assync it loses its reasons (at least wile being called through 'storage.Load'). 

I didn't saw no call to `storage.Load()` outside of `server.NewStorage()`  That is called through  `server.ListenAndServe`. 

Analyzing the `ListenAndServe`  it seems to keep the memory and the K8S storage in sync even without the  `s.syncStorage`, just needing the `secrets.OnChange` handler. 

The `server.getCA` gets the storage from memory/disk if no Secrets are passed through the options (not rancher case).
 If secrets are passed  (rancher case) it will try to get the CA from a secret and if it doesn't exist it will generate it and create the respective Secret.  

Every calls to the api invokes the listener.updateCert, that will check from the k8s storage and update it if necessary. After that  the `kubernetes/storage.Update`, that is triggered from any change to the Secret will do the update in memory.
 
 
 ---
 
If this is wright for making this code more accessible and easy to read/understand  I do think that we could remove the `go syncStorage()`   from the `storage.init` changing it to the "New if that function needs it". 

About setting the storage as ready I think that it can be done in `storage.init` or during the `storage.Load / storage.New`  (if new needs that function to be considered ready). 

Something on those lines: 

```
func Load(ctx context.Context, secrets v1controller.SecretController, namespace, name string, backing dynamiclistener.TLSStorage) dynamiclistener.TLSStorage {
	storage := &storage{...}
	storage.init(secrets)
	// This  line only if New needs the syncStorage to be considered as initialized. 
        storage.initialized = true 
	return storage
}


func New(ctx context.Context, core CoreGetter, namespace, name string, backing dynamiclistener.TLSStorage) dynamiclistener.TLSStorage {
	storage := &storage{...}

	// lazy init
	go func() {
		wait.PollImmediateUntilWithContext(ctx, time.Second, func(cxt context.Context) (bool, error) {
			if coreFactory := core(); coreFactory != nil {
				storage.init(coreFactory.Core().V1().Secret())
				go storage.syncStorage()
				return true, start.All(ctx, 5, coreFactory)
			}
			return false, nil
		})
	}()

	return storage
}

```

brandond:
> I didn't saw no call to storage.Load() outside of server.NewStorage() That is called through server.ListenAndServe.

K3s doesn't take that same path, we don't use `Load` or `ListenAndServe`. 
1. We create our own listener with `NewTCPListener` here:
    https://github.com/k3s-io/k3s/blob/master/pkg/cluster/https.go#L46-L46
2. We then up the file and kubernetes storage here:
    https://github.com/k3s-io/k3s/blob/master/pkg/cluster/https.go#L152-L160
3. We then pass the existing listener and storage into `NewListenerWithChain`, getting the wrapper listener and request handler back in return:
    https://github.com/k3s-io/k3s/blob/master/pkg/cluster/https.go#L61-L82
4. We start a goroutine to actually `Server` on the listener later:
    https://github.com/k3s-io/k3s/blob/master/pkg/cluster/https.go#L136-L141

felipe-colussi (changes requested):

storage/kubernetes/controller.go
```diff
@@ -160,7 +163,28 @@ func (s *storage) targetSecret() (*v1.Secret, error) {
 }
 
 func (s *storage) saveInK8s(secret *v1.Secret) (*v1.Secret, error) {
+
```
brandond, line 166:
```suggestion
```

storage/kubernetes/controller.go
```diff
@@ -160,7 +163,28 @@ func (s *storage) targetSecret() (*v1.Secret, error) {
...
+				secret, err := s.saveInK8s(secret)
+				if err != nil {
+					logrus.Errorf("failed to save TLS secret for %s/%s: %v", secret.Namespace, secret.Name, err)
+					return false, err
+				}
+				return true, nil
+			})
+
+			if err != nil {
+				logrus.Errorf("error while waiting for storage to initialized to save secret %s/%s: %v", secret.Namespace, secret.Name, err)
```
brandond, line 185: 1. The standard form for this when not returning is the more compact `if err := foo(); err != nil {`
2. Don't store the secret returned from the inner `saveInK8s` call; we're not going to do anything with it and it shadows the `secret` value from the outer scope.
3. Don't log the error from the inner `saveInK8s` twice, just return it and let it be logged as the err return from `Poll`
4. Log messages should be capitalized like a sentence.
```suggestion
			if err := wait.Poll(100*time.Millisecond, 15*time.Minute, func() (bool, error) {
				if !s.initComplete() {
					return false, nil
				}
				_, err := s.saveInK8s(secret)
				return true, err
			}); err != nil {
				logrus.Warnf("Failed to save TLS secret after controller init: %v", err)
```

brandond (changes requested):

storage/kubernetes/controller.go
vardhaman22: updated.

felipe-colussi (approved):
Tested on rancher: 
https://github.com/rancher/rancher/pull/42969
https://drone-pr.rancher.io/rancher/rancher/33380/4/2

and it worked.

brandond:
Updated my K3s PR for testing:
* https://github.com/k3s-io/k3s/pull/8401

PR #82, opened 2023-10-04
Bump wrangler to v2

brandond:
* Tracks package change from https://github.com/rancher/wrangler/pull/317


brandond:
already done in https://github.com/rancher/dynamiclistener/pull/97

PR #84, opened 2023-11-18
Add support for wildcard SANs

### LINKED ISSUE #7821: wildcard domain support with `--tls-san`
wildcard domain support with `--tls-san`, like `--tls-san=*.k8s.local --tls-san=*.xxx.com`

xdqi:
@brandond

This patch should fix https://github.com/k3s-io/k3s/issues/7821

Note: The original way of checking the validity of SAN[1] is not consistent with x509 logic of Go's standard library. So I adapted it.

[1]: Go never use CN field of a X509 certificate to validate, it use SAN only.

https://github.com/golang/go/blob/631a6c2abfb2cf7a877ea80f98c504fc4e0077be/src/crypto/x509/verify.go#L109

### DIFF
--- a/factory/gen.go
+++ b/factory/gen.go
@@ -12,7 +12,6 @@ import (
 	"encoding/pem"
 	"fmt"
 	"net"
-	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


factory/gen.go
```diff
@@ -244,7 +239,7 @@ func populateCN(secret *v1.Secret, cn ...string) *v1.Secret {
 		secret.Annotations = map[string]string{}
 	}
 	for _, cn := range cn {
-		if cnRegexp.MatchString(cn) {
+		if validHostnamePattern(cn) {
```
brandond, line 242: I'm not sure this is correct, as populateCN is also called to add CNs extracted from TLS handshakes and HTTP host headers. Since `validHostnamePattern` calls `validHostname(host, true)`, I believe this would allow unauthenticated callers to make dynamiclistener add wildcard SANs, if they spoof the correct value. I think wildcards should only be added if requested by the administrator.

brandond (changes requested):

brandond:
Done in https://github.com/rancher/dynamiclistener/pull/272

Issue #90, opened 2024-01-17
Customize the Expiration Days for Newly Signed CA Certificates


qkboy:
Now, the newly signed CA certificate is set to expire 10 years statically after its creation time and date.

There isn't suitable for live cluster.

We can add a new CATTLE_NEW_SIGNED_CA_EXPIRATION_YEARS environment variable to change this value.

PR #91, opened 2024-01-17
The CA expiration time can be defined from env, default is 100 years.

qkboy:
Issue #90


chenk008:
This is very useful for us.

But i think we can keep existing default value 10y.

chenk008:
cc @dereknola

PR #98, opened 2024-04-03
Add CI

tomleb:
## Problem & Solution

Couple of things:
- This repository is not running any tests when creating PRs. Let's make it do that.
-  I'm keeping fossa as is because it IS working. Let's remove it once we actually start the migration.
- I'm not running golangci-lint because there are non-trivial lint errors to fix that would require its own review.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,20 @@
+name: CI
+on:
+  pull_request:
+  push:
+    branches:
+    - 'master'
+    - 'release/*'
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      # https://github.com/actions/checkout/releases/tag/v4.1.1
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+    - name: Install Go
+      # https://github.com/actions/setup-go/releases/tag/v5.0.0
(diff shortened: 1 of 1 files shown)


maxsokolovsky (dismissed):
Just curious: how do we know what version of Go CI would choose? Is it determined by `setup-go` and its hash?

tomleb:
> Just curious: how do we know what version of Go CI would choose? Is it determined by `setup-go` and its hash?

So in the GHA file, I'm setting:

```yaml
      with:
        go-version-file: 'go.mod'
```

which means it will look at the version in the `go.mod` file to determine which Go version to pick which [in this case is 1.22](https://github.com/rancher/dynamiclistener/blob/master/go.mod#L3). I'm not 100% sure where it will download that Go version though, but we officially trust this github action so no concern here.

JonCrowther (approved):

tomleb:
I have removed the `Makefile` and the PR template for now to keep this PR simple. Since this is a repo who other contribute to, it's been suggested to seek feedback for adding a makefile.

I've changed the workflow so that it on runs on PR as well as when we push to new branches, but only for release branches. Seems to be what others are doing (eg: https://github.com/rancher/kube-api-auth/pull/23/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR33-R36, https://github.com/rancher/fleet/blob/main/.github/workflows/ci.yml#L9)

maxsokolovsky (approved):

JonCrowther (approved):

PR #118, opened 2024-07-16
Log panics from the server

ericpromislow:
Related to [#46002](https://github.com/rancher/rancher/issues/46002)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/rancher/dynamiclistener
 
-go 1.23.0
+go 1.22.0
+
+toolchain go1.22.1
 
 require (
 	github.com/rancher/wrangler/v3 v3.0.0
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/rancher/dynamiclistener
 
-go 1.22.0
-
-toolchain go1.22.1
+go 1.23.0
 
 require (
 	github.com/rancher/wrangler/v3 v3.0.0
--- a/server/server.go
+++ b/server/server.go
@@ -36,9 +36,32 @@ type ListenOpts struct {
 	BindHost          string
 	NoRedirect        bool
 	TLSListenerConfig dynamiclistener.Config
+
+	// Override legacy behavior where server logs written to the application's logrus object
+	// were dropped unless logrus was set to debug-level (such as by launching steve with '--debug').
+	// Setting this to true results in server logs appearing at an ERROR level.
+	DisplayServerLogs bool
 }
 
 func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.Handler, opts *ListenOpts) error {
+	writer := logrus.StandardLogger().WriterLevel(logrus.DebugLevel)
+	if opts.DisplayServerLogs {
+		// Create a new logger the writes to which are at the same level as it's
+		// configured at, with an "ERROR" prefix (as the server writes "un-levelled"
+		// text to the log stream it's handled).
+		// We aren't grabbing `logrus.StandardLogger()` because that gives a reference
+		// to the one global logger, and we don't want to change its logging level here.
+		logger := logrus.New()
+		logger.SetOutput(logrus.StandardLogger().Out)
+		logger.SetLevel(logrus.ErrorLevel)
+		writer = logger.WriterLevel(logrus.ErrorLevel)
+	}
+	// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
+	errorLog := log.New(writer, "", log.LstdFlags)
+	return listenAndServeWithLogger(ctx, httpsPort, httpPort, handler, opts, errorLog)
+}
+
+func listenAndServeWithLogger(ctx context.Context, httpsPort, httpPort int, handler http.Handler, opts *ListenOpts, errorLog *log.Logger) error {
 	if opts == nil {
 		opts = &ListenOpts{}
 	}
@@ -46,9 +69,6 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 	if opts.TLSListenerConfig.TLSConfig == nil {
 		opts.TLSListenerConfig.TLSConfig = &tls.Config{}
 	}
-
-	logger := logrus.StandardLogger()
-	errorLog := log.New(logger.WriterLevel(logrus.DebugLevel), "", log.LstdFlags)
 
 	if httpsPort > 0 {
 		tlsTCPListener, err := dynamiclistener.NewTCPListener(opts.BindHost, httpsPort)
--- a/server/server.go
+++ b/server/server.go
@@ -5,6 +5,7 @@ import (
 	"crypto"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"log"
 	"net"
@@ -46,18 +47,15 @@ type ListenOpts struct {
 func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.Handler, opts *ListenOpts) error {
 	logger := logrus.StandardLogger()
 	writer := logger.WriterLevel(logrus.DebugLevel)
+	if opts == nil {
+		opts = &ListenOpts{}
+	}
 	if opts.DisplayServerLogs {
 		writer = logger.WriterLevel(logrus.ErrorLevel)
 	}
 	// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
-	errorLog := log.New(writer, "", log.LstdFlags)
-	return listenAndServeWithLogger(ctx, httpsPort, httpPort, handler, opts, errorLog)
-}
 
-func listenAndServeWithLogger(ctx context.Context, httpsPort, httpPort int, handler http.Handler, opts *ListenOpts, errorLog *log.Logger) error {
-	if opts == nil {
-		opts = &ListenOpts{}
-	}
+	errorLog := log.New(writer, "", log.LstdFlags)
 
 	if opts.TLSListenerConfig.TLSConfig == nil {
 		opts.TLSListenerConfig.TLSConfig = &tls.Config{}
@@ -89,7 +87,7 @@ func listenAndServeWithLogger(ctx context.Context, httpsPort, httpPort int, hand
 		go func() {
 			logrus.Infof("Listening on %s:%d", opts.BindHost, httpsPort)
 			err := tlsServer.Serve(tlsTCPListener)
-			if err != http.ErrServerClosed && err != nil {
+			if !errors.Is(err, http.ErrServerClosed) && err != nil {
 				logrus.Fatalf("https server failed: %v", err)
 			}
 		}()
@@ -111,7 +109,7 @@ func listenAndServeWithLogger(ctx context.Context, httpsPort, httpPort int, hand
 		go func() {
 			logrus.Infof("Listening on %s:%d", opts.BindHost, httpPort)
 			err := httpServer.ListenAndServe()
-			if err != http.ErrServerClosed && err != nil {
+			if !errors.Is(err, http.ErrServerClosed) && err != nil {
 				logrus.Fatalf("http server failed: %v", err)
 			}
 		}()
--- a/server/server.go
+++ b/server/server.go
@@ -5,7 +5,6 @@ import (
 	"crypto"
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"fmt"
 	"log"
 	"net"
@@ -87,7 +86,7 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 		go func() {
 			logrus.Infof("Listening on %s:%d", opts.BindHost, httpsPort)
 			err := tlsServer.Serve(tlsTCPListener)
-			if !errors.Is(err, http.ErrServerClosed) && err != nil {
+			if err != http.ErrServerClosed && err != nil {
 				logrus.Fatalf("https server failed: %v", err)
 			}
 		}()
@@ -109,7 +108,7 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 		go func() {
 			logrus.Infof("Listening on %s:%d", opts.BindHost, httpPort)
 			err := httpServer.ListenAndServe()
-			if !errors.Is(err, http.ErrServerClosed) && err != nil {
+			if err != http.ErrServerClosed && err != nil {
 				logrus.Fatalf("http server failed: %v", err)
 			}
 		}()
--- a/server/server.go
+++ b/server/server.go
@@ -44,16 +44,9 @@ type ListenOpts struct {
 }
 
 func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.Handler, opts *ListenOpts) error {
-	writer := logrus.StandardLogger().WriterLevel(logrus.DebugLevel)
+	logger := logrus.StandardLogger()
+	writer := logger.WriterLevel(logrus.DebugLevel)
 	if opts.DisplayServerLogs {
-		// Create a new logger the writes to which are at the same level as it's
-		// configured at, with an "ERROR" prefix (as the server writes "un-levelled"
-		// text to the log stream it's handled).
-		// We aren't grabbing `logrus.StandardLogger()` because that gives a reference
-		// to the one global logger, and we don't want to change its logging level here.
-		logger := logrus.New()
-		logger.SetOutput(logrus.StandardLogger().Out)
-		logger.SetLevel(logrus.ErrorLevel)
 		writer = logger.WriterLevel(logrus.ErrorLevel)
 	}
 	// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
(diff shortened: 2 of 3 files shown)


server/server.go
```diff
@@ -48,7 +56,9 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 	}
 
 	logger := logrus.StandardLogger()
-	errorLog := log.New(logger.WriterLevel(logrus.DebugLevel), "", log.LstdFlags)
+	logger.SetLevel(logrus.DebugLevel)
```
MbolotSuse, line 59: I'm not sure that this solves the issue. From the issue, my understanding of the problem was that we were using logrus at debug level. Since logrus [defaults to info](https://pkg.go.dev/github.com/sirupsen/logrus#Logger), this effectively suppressed any output in default circumstances.

This PR does not seem to fix that - it's still at debug level and cannot be overriden by an argument in the options. Did I misunderstand the original issue?

MbolotSuse (changes requested):
In addition to the main comment, this should currently target `master`. Please change the branch when you get a moment.

server/server.go
ericpromislow: First change: log the panics from the server. I don't think they were even getting to the `dynamicListener` client.

ericpromislow: Second change: honor the same env vars that `webhook` does to put the logger in debug mode.

server/server.go
```diff
@@ -195,6 +201,12 @@ func newStorage(ctx context.Context, opts ListenOpts) dynamiclistener.TLSStorage
 
 func wrapHandler(handler http.Handler, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		defer func() {
+			if err := recover(); err != nil {
+				logrus.Errorf("Server panicked: %v", err)
+				http.Error(rw, fmt.Sprintf("Server panicked: %v", err), http.StatusInternalServerError)
```
tomleb, line 207: This seems like a big change and I'm not sure we want this.

[http.Server](https://pkg.go.dev/net/http#Server)'s already catches errors and logs them with its `ErrorLog` field (that's the field being set by the faulty `errorLog` that we're trying to fix.

I'm not sure we can send a `500` this way because the panic might occur after we start sending the headers (eg: 200).

I'd say let's just fix the error logging instead.

server/server.go
```diff
@@ -47,8 +48,13 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
...
 	}
 
+	logLevel := logrus.InfoLevel
+	// Use the same env-vars as webhook to log at the debug level
+	if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
+		logLevel = logrus.DebugLevel
+	}
 	logger := logrus.StandardLogger()
-	errorLog := log.New(logger.WriterLevel(logrus.DebugLevel), "", log.LstdFlags)
+	errorLog := log.New(logger.WriterLevel(logLevel), "", log.LstdFlags)
```
tomleb, line 57: I'm trying to get the following to run to verify the behavior, but nothing is printed no matter what log level is set:

```go
package main

import (
	"log"
	"os"

	"github.com/sirupsen/logrus"
)

func main() {
	logLevel := logrus.ErrorLevel
	// Use the same env-vars as webhook to log at the debug level
	if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
		logLevel = logrus.DebugLevel
	}
	logger := logrus.StandardLogger()
	errorLog := log.New(logger.WriterLevel(logLevel), "", log.LstdFlags)
	errorLog.Println("HI")
}
```

Does that run for you?

ericpromislow:
I backed out the change to the wrapped handler, and went back to `LogrusAdapter` change.

Then I made a change to `steve`, making this change:
```
func (a *apiServer) apiHandler(apiFunc APIFunc) http.Handler {
-       return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-               apiOp, ok := a.common(rw, req)
-               if ok {
-                       if apiFunc != nil {
-                               apiFunc(a.sf, apiOp)
-                       }
-                       a.server.Handle(apiOp)
-               }
+       return http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {
+               panic("Who let the dogs out")
        })
 }
```

I then started up `steve`, ran `curl -kL localhost:9080/v1` , and saw this in the steve logs:

```
2024/08/09 10:57:29 http2: panic serving [::1]:62420: Who let the dogs out
goroutine 1714 [running]:
net/http.(*http2serverConn).runHandler.func1()
	/opt/homebrew/Cellar/go/1.22.3/libexec/src/net/http/h2_bundle.go:6185 +0x13c
…
```

If I revert the changes to `dynamicListener` but not to Steve, the panic happens (because the curl call returns a curl error status with the same message as above:

```
curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
```

but there are no logs in Steve

ericpromislow:
Rebased against `release/v0.5`

MbolotSuse (changes requested):

server/server.go
```diff
@@ -47,8 +48,13 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 		opts.TLSListenerConfig.TLSConfig = &tls.Config{}
 	}
 
+	logLevel := logrus.InfoLevel
+	// Use the same env-vars as webhook to log at the debug level
+	if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
+		logLevel = logrus.DebugLevel
+	}
```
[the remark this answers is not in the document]
ericpromislow: Done

ericpromislow:
Playing some more, I think I see the problem.  Backing out all the changes except having the steve handler always panic, when I run `bin/steve --debug`, this is the Steve output:
```
…
INFO[0001] Watching metadata for rke.cattle.io/v1, Kind=RKEBootstrap 
INFO[0001] Watching metadata for /v1, Kind=ConfigMap    
DEBU[0002] 2024/08/13 13:27:24 http2: panic serving [::1]:65005: Aw crap. 
DEBU[0002] goroutine 3385 [running]:                    
DEBU[0002] net/http.(*http2serverConn).runHandler.func1() 
…
```

When I omit the `--debug` arg, the panic's don't show up. So it looks like `panic`s go through `log.Debug`, probably because all the output does.

See steve PR https://github.com/rancher/steve/pull/261 fixes this mismatch by keeping the logger and log-level in sync.

ericpromislow:
It targets master now.

ericpromislow:
I wrote an exploratory program with this code:
```
logger := logrus.StandardLogger()
	levels := []logrus.Level{logrus.DebugLevel, logrus.InfoLevel, logrus.ErrorLevel}
	names := []string{"Debug", "Info", "Error"}
	for i, x := range levels {
		for j, y := range levels {
			msg := fmt.Sprintf("[%d:%d]: Trying set-level %s, log to %s", i, j, names[i], names[j])
			// fmt.Println(msg)
			logrus.SetLevel(x)
			errorLog := log.New(logger.WriterLevel(y), "", log.LstdFlags)
			errorLog.Println(msg)
		}
	}
```

Output:
```
DEBU[0000] 2024/08/14 11:02:57 [0:0]: Trying set-level Debug, log to Debug 
INFO[0000] 2024/08/14 11:02:57 [0:1]: Trying set-level Debug, log to Info 
INFO[0000] 2024/08/14 11:02:57 [1:1]: Trying set-level Info, log to Info 
ERRO[0000] 2024/08/14 11:02:57 [1:2]: Trying set-level Info, log to Error 
ERRO[0000] 2024/08/14 11:02:57 [0:2]: Trying set-level Debug, log to Error 
ERRO[0000] 2024/08/14 11:02:57 [2:2]: Trying set-level Error, log to Error 
```

The old code was doing the equivalent of doing `set-level Info, log to Debug` so we never get any output from the server.

This fix gets all log output from the server, as we're shoehorning all its output into one log level.

server/server.go
```diff
@@ -48,7 +48,7 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 	}
 
 	logger := logrus.StandardLogger()
-	errorLog := log.New(logger.WriterLevel(logrus.DebugLevel), "", log.LstdFlags)
+	errorLog := log.New(logger.WriterLevel(logrus.GetLevel()), "", log.LstdFlags)
```
MbolotSuse, line 51: I like the simplicity of this solution, but I'm a bit concerned that this will always produce logs where before we weren't ever producing logs. Does this just log panics, or does this log a lot more?

server/server.go
```diff
@@ -36,6 +36,7 @@ type ListenOpts struct {
 	BindHost          string
 	NoRedirect        bool
 	TLSListenerConfig dynamiclistener.Config
+	DisplayServerLogs bool
```
tomleb, line 39: Would you mind adding a comment to describe what this option does?

server/server.go
```diff
@@ -47,8 +48,16 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 		opts.TLSListenerConfig.TLSConfig = &tls.Config{}
 	}
 
-	logger := logrus.StandardLogger()
-	errorLog := log.New(logger.WriterLevel(logrus.DebugLevel), "", log.LstdFlags)
+	writer := logrus.StandardLogger().WriterLevel(logrus.DebugLevel)
```
tomleb, line 51: nit: Maybe adding a comment about this being there for backwards compatibility reason

tomleb (commented):
Overall looks okay. `DisplayServerLogs` seems a tiny bit misleading since even if false, you could have panics being logged in the global logrus logger is at Debug loglevel. Not sure what a better name for this would be though. `ForceLog`? Heh

tomleb (approved):

server/server.go
```diff
@@ -47,8 +52,17 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
...
+	if opts.DisplayServerLogs {
+		// Create a new logger the writes to which are at the same level as it's
+		// configured at, with an "ERROR" prefix (as the server writes "un-levelled"
+		// text to the log stream it's handled).
+		logger := logrus.New()
```
MbolotSuse, line 60: By using `logrus.New()` rather than `logrus.StandardLogger()` I think this could discard changes made to the original standard logger by another application (e.g. rancher). Since we don't do this in the old case, I'm not sure why we do this here - can you elaborate on why these behave differently?

I would propose that this mirror the old side:

```suggestion
		logger := logrus.StandardLogger()
```

MbolotSuse (changes requested):
One comment on the use of `logrus.New()`

ericpromislow:
Not a great unit test because I couldn't re-use the server construction code in server.log. Instead I used similar code to show how the logs work.

ericpromislow:
And the "unit test" I wrote has a race condition when the logrus-wrapped `bytes.Buffer` object gets updated in the server running in a goroutine.  If I put the client in a goroutine and have the server run in the main goroutine, there's no race condition, but the `logBuf` is always empty.

I'm not sure about pursuing this to look for a solution -- this test is already not a unit test because it doesn't use the code in the actual package, but duplicates it. And I can't use `httptest` to mock the `http.Server` because `httptest.Server` doesn't take an `ErrorLog` field, the field we're trying to test.

I move that we pull out this testing code. I'm not sure how to implement a test on this -- when I started I modified `steve/pkg/server/handler/apiserver.go:apiHandler` to panic on every request, and verified that the panic's showed up. But I haven't landed on a good way to test this.

MbolotSuse (commented):
Couple of nits in addition to two major concerns:
- With the new option, we are still using logger.New() (which could cause us to discard options set on the standard logger).
- We are bumping golang minor versions, which could affect our module compatibility (in isolation bumping the go version is fine, but I think that should be a separate PR)

go.mod
```diff
@@ -1,11 +1,9 @@
 module github.com/rancher/dynamiclistener
 
-go 1.22.0
-
-toolchain go1.22.1
+go 1.23.0
```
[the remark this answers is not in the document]
ericpromislow: Because some other modules I've been reviewing lately were using go 1.23.0. Will revert.

server/server_test.go
```diff
@@ -0,0 +1,131 @@
...
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		fmt.Println("Error creating request:", err)
+		return
+	}
+	resp, err := client.Do(req)
+	assert.Error(err, "server should have panicked on request")
+	if resp != nil {
+		defer resp.Body.Close()
+		fmt.Println("Response status:", resp.Status)
```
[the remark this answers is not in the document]
ericpromislow: Sure, this was useful only during early development

server/server_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	logger := logrus.StandardLogger()
+	logger.SetOutput(safeWriter)
+	logger.SetLevel(logrus.ErrorLevel)
+	writer := logger.WriterLevel(logLevel)
+	errorLog := log.New(writer, "", log.LstdFlags)
+
+	err := listenAndServeWithLogger(ctx, httpsPort, httpPort, &handler, listenOpts, errorLog)
+	assert.Nil(err)
```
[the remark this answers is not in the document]
ericpromislow: makes sense

server/server_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	makeTheHttpRequest(assert, httpPort)
+	cancel()
+}
+
+func makeTheHttpRequest(assert *assertPkg.Assertions, port int) {
+	addr := fmt.Sprintf("127.0.0.1:%d", port)
```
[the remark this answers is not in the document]
ericpromislow: Sure.

server/server_test.go
```diff
@@ -0,0 +1,139 @@
...
+	totalTime := 0 * time.Millisecond
+	const maxWaitTime = 10 * time.Second
+	// Waiting for server to be ready..., max of maxWaitTime
+	for {
+		conn, err := net.Dial("tcp", addr)
+		if err == nil {
+			conn.Close()
+			break
+		} else if totalTime > maxWaitTime {
+			return fmt.Errorf("timed out waiting for the server to start after %d msec", totalTime/1e6)
```
MbolotSuse, line 117: Nit: the overall time spent waiting is useful, but I think the actual error that stopped this from working is more useful since it tells the user what caused it to wait so long.

MbolotSuse (dismissed):
One nit, but it's not necessary to fix. LGTM.

tomleb (approved):
LGTM.

Tbh, I would have preferred if the test would validate the interface of ListenAndServer, as in test that when ListenOpts{DisplayServerLogs: true} is given, it appropriately logs the panics. If we change the log setup in `ListenAndServe`, the current tests won't catch regressions.

ericpromislow:
Once I figured out how to avoid the race conditions, I realized there was no reason to break up `ListenAndServe` and pass in a separate logger object. The key also is always using the `logrus.StandardLogger()` global object and not creating a new logrus logger.

server/server.go
```diff
@@ -98,7 +109,7 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 		go func() {
 			logrus.Infof("Listening on %s:%d", opts.BindHost, httpPort)
 			err := httpServer.ListenAndServe()
-			if err != http.ErrServerClosed && err != nil {
+			if !errors.Is(err, http.ErrServerClosed) && err != nil {
```
MbolotSuse, line 112: I'd prefer that you not change this. I know that the new way makes more sense (and were you writing this for the first time I would say this is a good change), but the fatal case causes the server to exit early. This could cause the server to exit in new cases (like if something was returning a wrapped `http.ErrServerClosed`).

Since this is strictly unnecessary for your PR, I'd rather leave this as-is, we can clean this up in another PR (like something that we don't intend on back-porting).

MbolotSuse (changes requested):

server/server.go
ericpromislow: ok, I was getting deprecation warnings from a linter I'm using, but sure...

MbolotSuse (approved):

server/server_test.go
```diff
@@ -0,0 +1,134 @@
...
+	return makeTheHttpRequest(addr)
+}
+
+func makeTheHttpRequest(addr string) error {
+	url := fmt.Sprintf("%s://%s/", "http", addr)
+
+	waitTime := 10 * time.Millisecond
+	totalTime := 0 * time.Millisecond
+	const maxWaitTime = 10 * time.Second
+	// Waiting for server to be ready..., max of maxWaitTime
```
tomleb, line 105: Might be a nice feature to have a way to notify the readiness of the server. Like 

```go
ListenOptions{
  ReadyCh chan<- struct{}
}
```

anyway, not for this PR.

ericpromislow: I don't think the dynamiclistener code itself can do this.  `httpServer.ListenAndServe` doesn't take any channel arguments, and there aren't any on the object. It's called in a goroutine, and calls a nested one with `go c.serve(connCtx)`, so I don't see a non-polling way of recognizing when the server is ready.

PR #124, opened 2024-11-14
Implement versioning ADR

crobby:
https://github.com/rancher/rancher/issues/46905

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,21 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name : Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Create release on Github
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name : Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Create release on Github
+        run: |
+          gh --repo "${{ github.repository }}" release create ${{ github.ref_name }} --verify-tag --generate-notes
```
tomleb, line 19: Of course for the "example" case I picked the simplest that didn't require dealing with RCs.. Woops.

It would be great to detect whether the tag pushed is an RC and if so create a pre release instead of a release. This is done using the `--prerelease` flag like `gh --repo <repo> release --prerelease`. The tag pushed is `${{ github.ref_name }}` so you could detect `-rc` in it and add the flag if it is.

crobby: Is the check I added sufficient?  Or do we need to be more specific?

tomleb (approved):

.github/workflows/release.yaml
tomleb: That seems to work :+1:  Thanks!

joshmeranda (approved):

PR #184, opened 2025-07-03
Wrap Listener to also return dynamiclistener config

vitorsavian:
### Proposed Changes

- Use a wrapper for easy access to the `listener` config

### Why

- K3s need to regenerate leaf certs with `hot reload` without the need for restart
- the main idea is to also return the config that was used for creating the listener, since the best way to regenerate/reload the dynamic listener cert `on demand` is to use the `listener` that was created in `NewListenerWithChain` initial setup.

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -34,12 +34,18 @@ type SetFactory interface {
 	SetFactory(tls TLSFactory)
 }
 
+type ListenerWrapper struct {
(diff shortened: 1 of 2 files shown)


brandond:
We are actually the only users of NewListener/NewListenerWithChain - other rancher projects all just use server.ListenAndServe

PR #197, opened 2025-08-22
setting TLSError message as debug

marytlf:
## Problem
The Rancher webhook can generate a large volume of `http: TLS handshake error` logs when a client attempts to connect with a missing or invalid certificate. While these are often non-critical, they can overwhelm the logs, making it difficult to debug genuine issues. The current implementation doesn't provide a way to suppress these logs without impacting other important error messages. The community has requested a solution to reduce log noise caused by these specific errors, as detailed in issue #48484.
## Solution
This PR introduces a new environment variable, **IGNORE_TLS_HANDSHAKE_ERROR**, to the Rancher webhook. This variable, which defaults to `false`, gives operators control over the logging behavior of TLS handshake errors.

When **IGNORE_TLS_HANDSHAKE_ERROR** is set to `true`, `http: TLS handshake error `messages are downgraded from `error` to `debug` level. This means the messages will only appear in the logs if the **CATTLE_DEBUG** environment variable is also set to `true`. This approach provides a clear mechanism to either ignore these specific errors for a quieter log output or enable them for in-depth debugging.

## **Test Scenarios**
Scenario 1: `CATTLE_DEBUG = false, IGNORE_TLS_HANDSHAKE_ERROR = true`

Result: No TLS handshake error messages are logged. The messages are logged at the debug level but are not displayed because **CATTLE_DEBUG** is false.

Scenario 2: `CATTLE_DEBUG = false, IGNORE_TLS_HANDSHAKE_ERROR = false`

Result: No TLS handshake `error` messages are logged. The messages are logged at the `error` level, but the scenario does not produce the error.

Scenario 3:` CATTLE_DEBUG = true, IGNORE_TLS_HANDSHAKE_ERROR = true`

Result: TLS handshake `error` messages are logged at the `debug` level, as expected.

_`Log Example: level=debug msg="2025/08/21 18:03:25 http: TLS handshake error from 172.31.7.103:40900: EOF"`_

Scenario 4: `CATTLE_DEBUG = true, IGNORE_TLS_HANDSHAKE_ERROR = false`

Result: TLS handshake error messages are logged at the error level, preserving the default behavior.

_`Log Example: level=error msg="2025/08/21 18:17:38 http: TLS handshake error from 172.31.7.103:49312: EOF"`_

### How to reproduce and force the error:
NOTE: The error happens when using rke2, not k3s.
- Set the environment variables on rancher-webhook deployment `(kubectl -n cattle-system edit deploy rancher-webhook)`
- CATTLE_DEBUG and IGNORE_TLS_HANDSHAKE_ERROR
- Login on Rancher UI 
- Check the logs from rancher-webhook `(kubectl logs -l app=rancher-webhook -n cattle-system)`

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -54,7 +54,7 @@ type TLSErrorDebugger struct{}
 
 func (t *TLSErrorDebugger) Write(p []byte) (n int, err error) {
 	p = bytes.TrimSpace(p)
-	if bytes.Contains(p, TLSHandshakeError) {
+	if bytes.HasPrefix(p, TLSHandshakeError) {
 		logrus.Debug(string(p))
 	} else {
 		logrus.Error(string(p))
@@ -75,10 +75,10 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
 	var errorLog *log.Logger
 	if opts.IgnoreTLSHandshakeError {
 		debugWriter := &TLSErrorDebugger{}
-		errorLog = log.New(debugWriter, "", log.LstdFlags)
+		errorLog = log.New(debugWriter, "", 0)
 	} else {
 		// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
-		errorLog = log.New(writer, "", log.LstdFlags)
+		errorLog = log.New(writer, "", 0)
 	}
 
 	if opts.TLSListenerConfig.TLSConfig == nil {


crobby (commented):
I think this approach should work.  It would probably be fairly straightforward to include a unit test  of the tlsHandshakeErrorWriter part.  Not that it's super complex, but for the sake of completeness.

tomleb:
> The Rancher webhook can generate a large volume of http: TLS handshake error logs when a client attempts to connect with a missing or invalid certificate

The initial issue has to do with EOF error, not invalid certs. I'm wondering if we should be more granular in which error should be moved to debug. Should we check for general TLS handshake error, or should we verify that it's a TLS handshake error where the underlying error is EOF?

Any opinions? (cc: @brandond in case you're interested. We've discussed this in the past)

server/server_test.go
```diff
@@ -39,6 +40,82 @@ func (s *safeWriter) Write(p []byte) (n int, err error) {
 	return s.writer.Write(p)
 }
 
+func TestTLSHandshakeErrorWriter(t *testing.T) {
+	var mutex sync.Mutex
```
brandond, line 44: I don't see that you're using [`t.Parallel()`](https://pkg.go.dev/testing#T.Parallel) anywhere, nor are you starting any goroutines, so why would you need to add a mutex here to protect against concurrency?

server/server.go
```diff
@@ -40,8 +42,17 @@ type ListenOpts struct {
...
 	// were dropped unless logrus was set to debug-level (such as by launching steve with '--debug').
 	// Setting this to true results in server logs appearing at an ERROR level.
-	DisplayServerLogs bool
+	DisplayServerLogs       bool
+	IgnoreTLSHandshakeError bool
 }
+type tlsHandshakeErrorWriter struct {
+	defaultWriter io.Writer
+	debugWriter   io.Writer
+}
```
brandond, line 51: Do we really need a whole new type, another bool flag, and a bunch of tests just to do a `strings.Contains` on the error message?

brandond (changes requested):

brandond (commented):
What about something like this: https://go.dev/play/p/1kybUMvhYIx

marytlf:
@brandond @tomleb I addressed the changes Brad suggested but I kept the test, even if its just a simply check of a string "contains".
Also put the type definition close to its method, as Tom suggested.

server/server.go
```diff
@@ -65,12 +73,8 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
...
 
 	var errorLog *log.Logger
-
 	if opts.IgnoreTLSHandshakeError {
-		debugWriter := &tlsHandshakeErrorWriter{
-			defaultWriter: writer,
-			debugWriter:   logrus.StandardLogger().WriterLevel(logrus.DebugLevel),
-		}
+		debugWriter := &TLSErrorDebugger{}
 		errorLog = log.New(debugWriter, "", log.LstdFlags)
```
brandond: Nit: standard flags ends up creating duplicate timestamps, since standard logger adds a timestamp to the payload.

```golang
	l := log.New(&TLSErrorDebugger{}, "", log.LstdFlags)
	l.Print("test")
	l.Print("http: TLS handshake error: EOF")
```

```
time="2009-11-10T23:00:00Z" level=error msg="2009/11/10 23:00:00 test"
time="2009-11-10T23:00:00Z" level=error msg="2009/11/10 23:00:00 http: TLS handshake error: EOF"
```

Just use no flags:


```suggestion
		errorLog = log.New(&TLSErrorDebugger{}, "", 0)
```

brandond (changes requested):

server/server.go
```diff
@@ -52,9 +71,15 @@ func ListenAndServe(ctx context.Context, httpsPort, httpPort int, handler http.H
...
-	// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
 
-	errorLog := log.New(writer, "", log.LstdFlags)
+	var errorLog *log.Logger
+	if opts.IgnoreTLSHandshakeError {
+		debugWriter := &TLSErrorDebugger{}
+		errorLog = log.New(debugWriter, "", log.LstdFlags)
+	} else {
+		// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
+		errorLog = log.New(writer, "", log.LstdFlags)
```
brandond, line 81: Nit: standard flags ends up creating duplicate timestamps, since standard logger adds a timestamp to the payload.

```golang
	l := log.New(&TLSErrorDebugger{}, "", log.LstdFlags)
	l.Print("test")
	l.Print("http: TLS handshake error: EOF")
```

```
time="2009-11-10T23:00:00Z" level=error msg="2009/11/10 23:00:00 test"
time="2009-11-10T23:00:00Z" level=error msg="2009/11/10 23:00:00 http: TLS handshake error: EOF"
```

Just use no flags:

```suggestion
		errorLog = log.New(&TLSErrorDebugger{}, "", 0)
	} else {
		// Otherwise preserve legacy behaviour of displaying server logs only in debug mode.
		errorLog = log.New(writer, "", 0)
```

With this change I think you could also do bytes.HasPrefix instead of bytes.Contains, which is arguably more efficient - but I guess that would only work if the caller is consistent about not adding timestamp flags.

brandond (changes requested):

PR #202, opened 2025-08-27
Use concext cancellation to abort pending updates to backing Secret

brandond:
Fixes issue where multiple updates to Secret may all stack on top of each other if the apiserver is unavailable, causing excessive churn and CPU load once it returns.

This is a temporary workaround until https://github.com/rancher/rancher/issues/47333 is addressed, allowing context to be passed into the actual Create/Update calls.


galal-hussein (approved):

brandond:
Validated in k3s at 
* https://github.com/k3s-io/k3s/pull/12840

galal-hussein (approved):

PR #208, opened 2025-09-29
Avoid creating certs that violate Apple requirements for macOS 10.15

WesleyKlop:
## Context

The CA I use for my Kubernetes cluster has been valid for almost 2 years. This library used for generating the certificates for my k3s cluster generates certificates that use the CA NotBefore date as the NotBefore date for all leaf certificates. In the beginning this works fine, but after a while you run into this: [Apple documentation](https://support.apple.com/en-us/103769). My certificates are not "standards compliant" because they have a validity of more than 825 days. (ca NotBefore until moment of generation+1y)

## Changes

Changes generated certificates to have a NotBefore based on either the CA NotBefore or the current time. This prevents creation of certificates that are valid for too long making them return errors on platforms like MacOS.


brandond:
The certs managed by Dynamiclistener are not really meant to be compatible with CA/Browser Forum guidelines for certificate issuance, or any other vendor specific standards. They are meant to be used for internal communication between components in a Kubernetes cluster. They are compatible with Kubernetes clients, and that is sufficient for the goals of this project.

brandond:
This library is used to dynamically generate certificates that are shared across multiple pods or nodes. Whenever a new node/pod is created, the cert is updated to add addresses and hostnames. If we always set the NotBefore date to the current time, then we introduce a new dependency on strict time sync, where any node that has a clock that is even a few seconds behind will see the certificate as not yet valid.

The KB doc at https://support.apple.com/en-us/102028 says `This change will not affect certificates issued from user-added or administrator-added Root CAs` so I'm not sure why this would be affecting the Kubernetes certs, since these use a "user-added" root CA that is part of the kubeconfig. I guess they are saying that you would need to import the root CA into the system to bypass this check?

brandond:
> the CATTLE_NEW_SIGNED_CERT_EXPIRATION_DAYS currently does NOT generate a certificate that expires after $envvar days of being valid

No, it generates one that is valid that many days from the time it is signed. This is the expected behavior, and why I was suggesting not capping the total validity if someone requests a cert that is valid for an extended period of time. Most clients are NOT going to be on a Mac and there is no reason to break existing behavior just for the rare Mac user.

factory/cert_utils.go
```diff
@@ -98,13 +100,14 @@ func NewSignedCert(signer crypto.Signer, caCert *x509.Certificate, caKey crypto.
 		}
 	}
 
+	notBefore := cert.CalculateNotBeforeNow(caCert)
```
brandond, line 103:
```suggestion
	notBefore := cert.CalculateNotBefore(caCert)
```

cert/validity_test.go
```diff
@@ -0,0 +1,75 @@
...
+			ca: &x509.Certificate{
+				NotBefore: baseTime,
+			},
+			now:      baseTime,
+			expected: baseTime,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
brandond, line 67: Set the package clock var to your NewFakeClock() instance here

brandond (changes requested):
I guess I'm onboard with this. I have some nits about the implementation and tests though. Thanks for sticking with me!

brandond (approved):

Issue #122, opened 2024-10-09
Can we support wildcard tls cert for k3s?


mogliang:
Hi all,

We are using Azure relay to connect k3s cluster, the relay endpoint is like <guid>.myrelaydomain.com, guid is random generated at runtime. To make apiserver pass tls validation, we need add wildcard tls-san *.myrelaydomain.com, however, k3s don't like it

![image](https://github.com/user-attachments/assets/6382f014-ff68-4012-af8e-182995633ec2)

I see the regex reject wildcard cert, any specific reason for this? can we allow wildcard cert?

https://github.com/rancher/dynamiclistener/blame/3fc2b04ceda3f071d8862a3b12f56f058cb9b6e3/factory/gen.go#L32

mogliang:
add reference 
https://github.com/k3s-io/k3s/issues/7821

PR #272, opened 2026-05-08
Add support for wildcard SANs

Eshaan-Lumba:
## Summary

Adds support for wildcard SANs (e.g. `*.example.com`) in dynamiclistener. Wildcards may be configured via `Config.SANs` and follow RFC 6125 single-label matching semantics.

## Changes

- `factory/gen.go`: relax `cnRegexp` to accept RFC 6125 leading-label wildcards.
- `factory/gen.go`: extend `getAnnotationKey` to escape `*` (same encoding path as `:` for IPv6 — replace, then truncate-and-hash).
- `factory/gen.go`: `NeedsUpdate` now recognizes that an existing wildcard SAN covers matching subdomains per RFC 6125, avoiding redundant cert regeneration when SNI brings in a subdomain already covered by a configured wildcard.
- `listener.go`: filter wildcards from runtime sources (TLS SNI, TCP `LocalAddr`, HTTP `Host`) before they reach `updateCert`. Wildcards from `Config.SANs` (admin-supplied at startup) flow through unchanged.

## Trust model

| Source | Wildcards |
|---|---|
| `Config.SANs` (admin config) | accepted |
| `Config.FilterCN` (integrator-supplied) | accepted; treated as part of the integrator-controlled trust boundary |
| TLS SNI / TCP `LocalAddr` / HTTP `Host` (runtime, attacker-controlled) | rejected at the listener layer before `updateCert` is called |

Runtime-discovered SANs cannot introduce wildcards into the cert under this model. `populateCN` still validates via the relaxed `cnRegexp` as a second line of defense.

## Tests

- `factory/gen_test.go` (new): regex acceptance, annotation key encoding, `NeedsUpdate` wildcard coverage, `isCoveredByWildcard` helper, `Renew`/`Regenerate`/`Merge` wildcard preservation, `MaxSANs` interaction, long-hostname + wildcard encoding, IPv6 + wildcard coexistence, mixed wildcard + specific in a single `AddCN` call.
- `listener_test.go` (additions): `isWildcardSAN` helper, wildcard rejection from TLS SNI, admin wildcard from `Config.SANs` is accepted, no regen for covered subdomain (verifies `NeedsUpdate` cover semantics reach the cert), `Filter` callback behavior documented.

End-to-end validation against k3s + native Linux also performed locally:

1. Cert SAN list contains the admin-configured `*.edgeconnectivity.local`.
2. TLS handshake to `abc123.edgeconnectivity.local` succeeds via RFC 6125 wildcard match.
3. SNI for a covered subdomain does NOT trigger cert regeneration (fingerprint unchanged).
4. SNI containing a wildcard from an unauthenticated source is NOT added to the cert.

## Backward compatibility

None affected. The relaxed `cnRegexp` is a strict superset of the previous pattern; existing valid CNs still validate. Wildcard CNs were previously rejected outright, so no existing deployments could have stored them.

## References

- rancher/dynamiclistener#122
- k3s-io/k3s#7821

### DIFF
--- a/factory/gen.go
+++ b/factory/gen.go
@@ -29,7 +29,7 @@ const (
 )
 
 var (
-	cnRegexp = regexp.MustCompile("^([A-Za-z0-9:][-A-Za-z0-9_.:]*)?[A-Za-z0-9:]$")
+	cnRegexp = regexp.MustCompile(`^(\*\.)?([A-Za-z0-9:][-A-Za-z0-9_.:]*)?[A-Za-z0-9:]$`)
 )
 
 type TLS struct {
--- a/factory/gen.go
+++ b/factory/gen.go
@@ -268,22 +268,49 @@ func IsStatic(secret *v1.Secret) bool {
 
 // NeedsUpdate returns true if any of the CNs are not currently present on the
 // secret's Certificate, as recorded in the cnPrefix annotations. It will return
-// false if all requested CNs are already present, or if maxSANs is non-zero and has
+// false if all requested CNs are already present (either explicitly, or covered
+// by an existing wildcard SAN per RFC 6125), or if maxSANs is non-zero and has
 // been exceeded.
 func NeedsUpdate(maxSANs int, secret *v1.Secret, cn ...string) bool {
 	if secret == nil {
 		return true
 	}
-
+	existingCNs := cns(secret)
 	for _, cn := range cn {
-		if secret.Annotations[getAnnotationKey(cn)] == "" {
-			if maxSANs > 0 && len(cns(secret)) >= maxSANs {
-				return false
-			}
+		if secret.Annotations[getAnnotationKey(cn)] != "" {
+			continue
+		}
+		if isCoveredByWildcard(cn, existingCNs) {
+			continue
+		}
+		if maxSANs > 0 && len(existingCNs) >= maxSANs {
+			return false
+		}
+		return true
+	}
+	return false
+}
+
+// isCoveredByWildcard reports whether cn is matched by any "*.parent" entry in
+// existing, per RFC 6125 single-label leftmost-wildcard semantics.
+//
+//	"*.example.com" covers "foo.example.com" but NOT "a.b.example.com" and NOT "example.com".
+//
+// A wildcard cn is never considered covered by another wildcard.
+func isCoveredByWildcard(cn string, existing []string) bool {
+	if strings.HasPrefix(cn, "*.") {
+		return false
+	}
+	dot := strings.IndexByte(cn, '.')
+	if dot < 1 {
+		return false
+	}
+	parent := "*" + cn[dot:]
+	for _, e := range existing {
+		if e == parent {
 			return true
 		}
 	}
-
 	return false
 }
 
--- a/factory/gen.go
+++ b/factory/gen.go
@@ -348,11 +348,12 @@ func NewPrivateKey() (crypto.Signer, error) {
 func getAnnotationKey(cn string) string {
 	cn = cnPrefix + cn
 	cnLen := len(cn)
-	if cnLen < 64 && !strings.ContainsRune(cn, ':') {
+	if cnLen < 64 && !strings.ContainsRune(cn, ':') && !strings.ContainsRune(cn, '*') {
 		return cn
 	}
 	digest := sha256.Sum256([]byte(cn))
 	cn = strings.ReplaceAll(cn, ":", "_")
+	cn = strings.ReplaceAll(cn, "*", "_")
 	if cnLen > 56 {
 		cnLen = 56
 	}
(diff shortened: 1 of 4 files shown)


Eshaan-Lumba:
@brandond would you be able to review this PR please?

brandond (approved):
this looks good to me, but someone from the frameworks team will also need to review.

Eshaan-Lumba:
Hi @crobby, brandond approved this PR but it still needs a review from the frameworks team. Would you be able to take a look please? Thanks!

crobby:
> Hi @crobby, brandond approved this PR but it still needs a review from the frameworks team. Would you be able to take a look please? Thanks!

Thanks for the contribution.  I have this on my list of things to review.

crobby:
QA Test Notes

  What changed

dynamiclistener now accepts wildcard SANs (e.g. *.example.com) when passed via admin config (--tls-san). Wildcards from untrusted runtime sources (TLS SNI, TCP address, HTTP Host header) are rejected.

  Prerequisites

  - k3s build that includes this dynamiclistener version
  - DNS or /etc/hosts entries pointing *.example.com (or your test domain) at the k3s node IP
  - openssl CLI for cert inspection

  Proposed Test Cases, add or subtract as needed

  1. Admin wildcard SAN gets added to cert
  # Start k3s with wildcard SAN
  k3s server --tls-san "*.example.com"

  # Verify cert contains wildcard
  echo | openssl s_client -connect <node-ip>:6443 -servername foo.example.com 2>/dev/null \
    | openssl x509 -noout -text | grep -A1 "Subject Alternative Name"
  # Expect: DNS:*.example.com in SAN list

  2. TLS handshake succeeds for covered subdomains
  # Should succeed — foo.example.com covered by *.example.com
  curl -k https://foo.example.com:6443/healthz
  curl -k https://bar.example.com:6443/healthz

  3. No cert regeneration for covered subdomains
  # Get initial cert fingerprint
  FINGER1=$(echo | openssl s_client -connect <node-ip>:6443 -servername test1.example.com 2>/dev/null \
    | openssl x509 -noout -fingerprint)

  # Hit with different covered subdomain
  curl -k https://test2.example.com:6443/healthz

  # Fingerprint should be unchanged
  FINGER2=$(echo | openssl s_client -connect <node-ip>:6443 -servername test2.example.com 2>/dev/null \
    | openssl x509 -noout -fingerprint)
    
  [ "$FINGER1" = "$FINGER2" ] && echo "PASS: no regen" || echo "FAIL: cert regenerated"
  
  4. Multi-label subdomain NOT covered (expected behavior per RFC 6125)
  # a.b.example.com is NOT covered by *.example.com
  # This should trigger cert regen and add a.b.example.com as explicit SAN
  curl -k https://a.b.example.com:6443/healthz
  
  echo | openssl s_client -connect <node-ip>:6443 -servername a.b.example.com 2>/dev/null \
    | openssl x509 -noout -text | grep "a.b.example.com"
  # Expect: a.b.example.com added as separate SAN

  5. Wildcard from SNI rejected (security)
  # Craft TLS ClientHello with wildcard ServerName — should NOT appear in cert
  openssl s_client -connect <node-ip>:6443 -servername "*.evil.com" </dev/null 2>/dev/null

  echo | openssl s_client -connect <node-ip>:6443 2>/dev/null \
    | openssl x509 -noout -text | grep "evil"
  # Expect: NO evil.com entry in cert
  
  6. Multiple --tls-san flags (wildcard + specific) coexist
  k3s server --tls-san "*.example.com" --tls-san "other.org"

  # Both should appear
  echo | openssl s_client -connect <node-ip>:6443 2>/dev/null \
    | openssl x509 -noout -text | grep -E '\*\.example\.com|other\.org'

  7. Backward compat — non-wildcard SANs still work
  # Plain hostname, IP, FQDN should all work exactly as before
  k3s server --tls-san "myhost.local" --tls-san "10.0.0.5"

  8. Cert renewal preserves wildcard
  # After forcing cert rotation (delete serving cert secret, restart k3s),
  # wildcard should still be present
  kubectl delete secret -n kube-system serving-cert  # or equivalent
  # restart k3s
  # re-check SAN list includes *.example.com

  What to watch for

  - Cert churn in logs ("dynamiclistener.*failed to update cert") — should NOT happen for covered subdomains
  - MaxSANs behavior unchanged — if SAN limit reached, new SANs (including wildcards) stop being added
  - No regression on IPv6 SAN handling

crobby (approved):
I think this looks good.  Thanks.  Not sure when it will merge as we need to coordinate with other teams.

brandond:
@crobby I am not sure why we would need extensive coordination for this. Other teams don't need to use this feature if they don't want it. It doesn't change anything about how dynamiclistener functions for the rancher or webhook use cases.

PR #301, opened 2026-06-17
Add FilterExisting to dynamiclistener Config

tomleb:
## Summary

- Adds `FilterExisting bool` to `Config` alongside the existing `FilterCN`
- When `FilterExisting` is `false` (the default), all existing CNs are preserved — identical to the previous behaviour (fully backwards compatible)
- When `FilterExisting` is `true`, `FilterCN` is also applied to the set of CNs already recorded in the secret's `listener.cattle.io/cn-*` annotations during every cert operation (`AddCN`, `Merge`, `Renew`, `Regenerate`, cert generation). Any existing CN that `FilterCN` would reject is pruned from the certificate on the next write — no delete-and-regenerate required
- A single filter callback (`FilterCN`) now governs both what can be added and, when `FilterExisting` is set, what is kept from existing entries
- `Merge` is updated to call `generateCert` when the winning cert contains stale CNs, ensuring they are pruned on the next storage sync rather than silently propagated (important for HA deployments)
- `pruneAnnotations` is moved to after `populateCN` in `generateCert` so CNs added by the caller but rejected by `FilterCN` are also stripped from annotations, keeping the X.509 SAN list and annotation set in sync
- `FilterExisting` is wired through `NewListenerWithChain`; SANs pre-seeded in `Config.SANs` are always preserved via `allowDefaultSANs` regardless of the filter

## Test plan

- [ ] Unit tests added for `hasStaleCNs`, `pruneAnnotations`, `Merge` (stale in target, stale in additional, no stale, `FilterExisting=false`, static target), `Renew`, `Regenerate`, and `AddCN`
- [ ] `go test ./...` passes

### DIFF
--- a/factory/gen.go
+++ b/factory/gen.go
@@ -37,10 +37,19 @@ type TLS struct {
 	CAKey               crypto.Signer
 	CN                  string
 	Organization        []string
-	FilterCN            func(...string) []string
+	FilterCN        func(...string) []string
+	// FilterExisting, when true, applies FilterCN to the set of CNs already
+	// recorded on the secret (via cnPrefix annotations) inside Merge, AddCN,
+	// Renew, Regenerate, and cert generation. Any existing CN that FilterCN
+	// would reject is pruned from the cert on the next write. false (default)
+	// preserves all existing CNs as before.
+	FilterExisting bool
 	ExpirationDaysCheck int
 }
 
+// cns returns the raw CN set recorded on the secret via cnPrefix
+// annotations. No filtering is applied. Used for raw-count checks (e.g.
+// the MaxSANs cap in NeedsUpdate) where the literal stored count matters.
 func cns(secret *v1.Secret) (cns []string) {
 	if secret == nil {
 		return nil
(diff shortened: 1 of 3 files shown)


factory/gen.go
```diff
@@ -37,10 +37,19 @@ type TLS struct {
...
 	CN                  string
 	Organization        []string
-	FilterCN            func(...string) []string
+	FilterCN        func(...string) []string
+	// PruneExistingCN, when true, applies FilterCN to the set of CNs already
+	// recorded on the secret (via cnPrefix annotations) inside Merge, AddCN,
+	// Renew, Regenerate, and cert generation. Any existing CN that FilterCN
+	// would reject is pruned from the cert on the next write. false (default)
+	// preserves all existing CNs as before.
+	PruneExistingCN bool
```
brandond: nit: this might be better named `FilterExisting`, since it allows the FilterCN function to see the existing+proposed CNs, instead of just the proposed CNs necessary to satisfy the in-flight request.

brandond (changes requested):
Have we tested this with multiple pods or hosts running against the same secret? My concern is that they will fight over the cert, since they each see that the peer's IP is not in the filtered list, and cycle back and forth adding their IP and removing their peers' IPs.

tomleb:
@brandond I have lightly tested this case already by running Rancher in HA. That use case works, I haven't seen a lot of fighting. The fighting should converge _eventually_ since all 3 pods will have their caches updated so they'll have the same target pod IPs.

One more use case I should test is upgrade case, because in that case, an old pod will NOT have the filterExisting field set to true, so until it is rolled out, then it might keep adding the pod IP.

Do you have any other use cases in mind that I could test? I'll try to think about other as well.

brandond:
K3s does something similar with the FilterCN function, but it uses node IPs - and has the same problem with thrashing until the new IPs and names are visible to all peers.

Have we done any profiling to ensure that the new more in-depth CN checks are not in the hot path anywhere?

tomleb:
> Have we done any profiling to ensure that the new more in-depth CN checks are not in the hot path anywhere?

I haven't done any profiling for this. What is considered the hot path here? When answering requests from users? I'm willing to give a go at profiling, is this something you've already done before for dynamiclistener? DO you already have some kind of setup to do profiling / testing here? Or should I just run Rancher HA and collect profiles?

brandond (approved):

rohitsuse (approved):

Issue #311, opened 2026-07-09
Dynamiclistener should use atomic writes for file-backed secret


brandond:
See upstream issue at: 
* https://github.com/k3s-io/k3s/issues/14220

Dynamiclistener should treat empty or corrupt file as nonexistent, and should write the file atomically to avoid corrupting it if the process is terminated mid-write.

https://github.com/rancher/dynamiclistener/blob/936962a7f3a10576c7b10099695a7a3775312276/storage/file/file.go#L34-L42

If the process is terminated between the time the file is opened, and the Encode function returns, the file may be left empty or only partially written.

Looking at `github.com/google/renameio` package to handle atomic writes for k3s and flannel.

PR #312, opened 2026-07-09
[storage/file] Use atomic writes; treat empty file as nonexistent

brandond:
Fixes issue where interrupted writes may leave the secret-backing file empty.

#### Linked Issues
* https://github.com/rancher/dynamiclistener/issues/311
* https://github.com/k3s-io/k3s/issues/14220


ericpromislow (approved):
LG

PR #313, opened 2026-07-10
Fall back to os.WriteFile on windows

brandond:
As per https://github.com/google/renameio/blob/v2.0.2/README.md#windows-support is is not possible to do atomic file replacement on Windows, and renameio doesn't export any functions on Windows.

### DIFF
--- a/storage/file/file.go
+++ b/storage/file/file.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"os"
 
-	"github.com/google/renameio/v2"
 	"github.com/rancher/dynamiclistener"
 	v1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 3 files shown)


vitorsavian (approved):

neiltalap:
This is not handling the file safely and correctly on Windows (I will use MoveFileExW as the most robust implementation). I will need to make similar changes here and in the k3s and flannel repositories.

Will reference the PRs where these changes are also being implemented by me:
https://github.com/flannel-io/flannel/pull/2486
https://github.com/k3s-io/k3s/pull/14342

I will create the needed changes in a separate PR in this repository as well.

brandond:
If the folks at Google say that it can't be done safely, I would tend to trust them. Please see if you can contribute your code to the renameio project, I don't think it belongs here.

PR #314, opened 2026-07-11
Add Windows support for writeFile

neiltalap:
Replace `os.WriteFile` with a proper atomic write using a temp file and `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING`, matching the atomic-rename semantics that the Unix implementation gets from `renameio`.


neiltalap:
This PR has been made in relation to another PR in the k3s repository https://github.com/k3s-io/k3s/pull/14342#issuecomment-4940069914

And in relation to a PR in this repository https://github.com/rancher/dynamiclistener/pull/313

brandond:
The google maintainer of renameio seem to believe that it is not possible to reliably write files atomically on Windows. See the reference to the renamio docs.

As I have said in other places, we DO NOT want to maintain our own code for atomic writes.

If you disagree with the renameio maintainers and believe there is a way to safely do this on Windows, please see if you can contribute it upstream to the renameio project.

neiltalap:
> The google maintainer of renameio seem to believe that it is not possible to reliably write files atomically on Windows. See the reference to the renamio docs.
> 
> As I have said in other places, we DO NOT want to maintain our own code for atomic writes.
> 
> If you disagree with the renameio maintainers and believe there is a way to safely do this on Windows, please see if you can contribute it upstream to the renameio project.

I will have to write a windows implementation for them then. 

In the mean time, do you suggest a simple os write file in k3s part of things?

neiltalap:
I think I will keep this open for now. If the maintainer of renameio does not like my rebuttal, then we will just use the simple write file operation.

PR #315, opened 2026-07-14
[storage/kubernetes] Guard queuedSecret with a mutex to fix data race

somaz94:
`storage.queuedSecret` was written from three goroutines (the `Update` caller,
the secret watch, and the initial sync in `init`) and read by the workqueue
processor's `update`, with no synchronization, so certificate updates could
race (#247).

This adds a mutex to the `storage` struct and routes every `queuedSecret`
access through `setQueuedSecret` / `getQueuedSecret`. `update` snapshots the
queued secret once before the conflict-retry loop, so the lock is never held
across an apiserver call and a concurrent write cannot swap the secret
mid-retry.

Scope: this fixes the data race only. The logical lost-update race (a watch
event overwriting a freshly generated cert) noted in #247 is left for a
follow-up.

Validation (local):
- Added `storage/kubernetes/controller_test.go` reproducing the race.
  `go test -race ./storage/kubernetes/` reports the race on the current code
  (controller.go:82 vs :247) and passes with the fix.
- `go vet ./...`, `go build ./...`, `go test -race -cover ./...` all pass.

Fixes #247

### DIFF
--- a/storage/kubernetes/controller.go
+++ b/storage/kubernetes/controller.go
@@ -35,7 +35,7 @@ type storage struct {
 
 	// mu guards queuedSecret, which is set from the Update caller and the
 	// secret watch goroutine, and read by the workqueue processor.
-	mu           sync.Mutex
+	mu           sync.RWMutex
 	queuedSecret *v1.Secret
 }
 
@@ -46,8 +46,8 @@ func (s *storage) setQueuedSecret(secret *v1.Secret) {
 }
 
 func (s *storage) getQueuedSecret() *v1.Secret {
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	return s.queuedSecret
 }
 
--- a/storage/kubernetes/controller.go
+++ b/storage/kubernetes/controller.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"context"
 	"maps"
+	"sync"
 	"time"
 
 	"github.com/rancher/dynamiclistener"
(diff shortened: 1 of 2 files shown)


storage/kubernetes/controller.go
```diff
@@ -31,7 +32,23 @@ type storage struct {
...
+
+func (s *storage) setQueuedSecret(secret *v1.Secret) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.queuedSecret = secret
+}
+
+func (s *storage) getQueuedSecret() *v1.Secret {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
brandond: nit: make it an RWMutex so that multiple threads can read concurrently

```suggestion
	// mu guards queuedSecret, which is set from the Update caller and the
	// secret watch goroutine, and read by the workqueue processor.
	mu           sync.RWMutex
	queuedSecret *v1.Secret
}

func (s *storage) setQueuedSecret(secret *v1.Secret) {
	s.mu.Lock()
	defer s.mu.Unlock()
	s.queuedSecret = secret
}

func (s *storage) getQueuedSecret() *v1.Secret {
	s.mu.RLock()
	defer s.mu.RUnlock()
```

brandond (changes requested):

brandond:
lgtm with the noted nit on RWMutex. Also, I suspect that this only fixes a small subset of the actual occurrences of the noted issue; without properly handling overlapping local writes and writes from other peers updating the secret, this fix is pretty marginal.

somaz94:
Applied the RWMutex nit in `37a6028` — `getQueuedSecret` now takes a read lock so the workqueue processor (and any future concurrent readers) can proceed in parallel; the setter keeps the write lock. `go test -race` still passes.

On the scope concern — agreed, this is deliberately narrow. It only closes the `-race`-detected data race on the in-memory `queuedSecret` field, because the broader cases you mention are already covered by the existing path:

- **Overlapping local writes** (listener adding SANs, cert regen) accumulate through `saveInK8s` → `tls.Merge`, which merges the queued secret against *both* the backing store and the live Kubernetes secret rather than blind-overwriting.
- **Peer writes** to the same secret come back in via the field-selected watch (re-queued → merged), and any write conflict is absorbed by the existing `retry.OnError(isConflictOrAlreadyExists)` loop.

So this isn't trying to redesign the coordination — just to make the existing queue-based path race-free. Happy to open a follow-up if you'd like the peer/merge semantics hardened beyond that, but it felt like a larger change than this fix warrants.

brandond (approved):

ericpromislow (approved):
Looks fine. Agreed it solves a specific problem, but I don't see it blocking any further developments.

PR #332, opened 2026-08-13
Adopt self-contained per-branch renovate rules, trim unsupported branches

crobby:
Part of rancher/rancher#56562.

**Reworked per review feedback** — no longer ties dynamiclistener's Renovate config to `rancher/renovate-config`'s per-rancher-line presets (`rancher-2.11.json` etc.), since this repo is also consumed by k3s and other non-rancher projects. `.github/renovate.json` now carries local rules keyed to dynamiclistener's own declared Kubernetes support range (`VERSION.md`), not a rancher release line:

- `release/v0.8` (Kubernetes v1.27 - v1.35) -> `k8s.io/*` capped at `<0.36`
- `release/v0.7` (Kubernetes v1.27 - v1.34) -> `k8s.io/*` capped at `<0.35`
- `release/v0.6` (Kubernetes v1.27 - v1.32) -> `k8s.io/*` capped at `<0.33`
- `main` (v1.27+, open-ended) -> no ceiling

`k8s.io/utils`, `k8s.io/kube-openapi`, and `k8s.io/klog/v2` are excluded from the ceiling groups since they don't follow the client-go `v0.<minor>` versioning scheme the ceiling assumes.

The top-level `rancher/renovate-config#release` extend stays — that's generic automerge/stability policy, not tied to a rancher release line, so it doesn't reintroduce the k3s-compatibility concern. It already provides the 7-day stability window and automerge for patch/minor + security updates. Kept alongside it: the existing guards against major bumps and Go/`golang.org/x/tools` minor bumps on release branches, and the digest-freeze rule that keeps the pinned preset ref from drifting on release branches.

`release/v0.5`, `release/v0.4`, and `release/v0.3` have no active consumer, so they're dropped from `baseBranchPatterns` entirely, and the corresponding rows are trimmed from `VERSION.md` to match.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,11 +4,9 @@
   ],
   "baseBranchPatterns": [
     "main",
+    "release/v0.8",
     "release/v0.7",
-    "release/v0.6",
-    "release/v0.5",
-    "release/v0.4",
-    "release/v0.3"
+    "release/v0.6"
   ],
   "prHourlyLimit": 2,
   "enabledManagers": [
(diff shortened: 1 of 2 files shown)


tomleb (commented):
Hummm Im not sure if we want to tie dynamiclistener with rancher this way since it's also used in other projects like k3s. Thoughts? Maybe it's fine for this one to have its own rules. 🤔

crobby:
> Hummm Im not sure if we want to tie dynamiclistener with rancher this way since it's also used in other projects like k3s. Thoughts? Maybe it's fine for this one to have its own rules. 🤔

You might be right on that one.  I'll make some changes and try to make sure we have a comprehensive set of local rules.

crobby:
Pushed a rework: dropped the `rancher/renovate-config` per-line `extends` entirely and replaced them with local rules keyed to dynamiclistener's own `VERSION.md` Kubernetes support range instead of a rancher release line. See the updated PR description for details.

rohitsuse (approved):

tomleb (approved):

